PR #10, opened 2023-08-22
Feat(connext): Add support for multisend ETH bridge

jimobrien:
When bridging ETH, Connext first wraps the eth. These transactions are not sent to the Connext bridge contract, but rather a multisend contract that first wraps the eth, then makes a delegate call to the bridge contract. 

Read more here:
https://github.com/connext/monorepo/issues/4218
https://github.com/connext/monorepo/issues/4218


packages/connext/src/Connext.ts
```diff
@@ -1,73 +1,100 @@
...
     contractAddress,
     tokenAddress,
     amount,
     recipient,
-  } = bridge
+  } = bridge;
+
+  const defaultContractAddress = ConnextContract[sourceChainId];
+  const destinationDomain = chainIdToDomain(destinationChainId);
+  const requiresWrapperMultisend = tokenAddress === ETH_TOKEN_ADDRESS;
```
jonathandiep, line 61: `ETH_TOKEN_ADDRESS` is set to the zero address, but I'm curious about why? Will the `tokenAddress` bridge param need to use zero address for Connext WETH bridging?

jimobrien: Yep basically that 

https://raw.githubusercontent.com/connext/chaindata/main/crossChain.json

```
"assetId": {
      "0x0000000000000000000000000000000000000000": {
        "symbol": "ETH",
        "mainnetEquivalent": "0x0000000000000000000000000000000000000000",
        "decimals": 18,
        "coingeckoId": "ethereum"
      },
      ...
},

PR #11, opened 2023-08-22
Feat(template): init template

Quazia:
Wasn't sure where to put the template but I landed on adding it to the packages root and then excluding it from all the turbo commands so we wouldn't constantly be failing tests/builds because of it. Let me know if you think there's a better approach certainly open to it. Also added some comments/read-me to make it super easy and obvious how to implement a plugin.


jimobrien (approved):
🔥 🔥 🔥 🔥 🔥 🔥 🔥

PR #13, opened 2023-08-24
Feat(arbitrum): Arbitrum bridge

Quazia:
Still needs some more testing - will edit and give a more complete PR description after I wrap up testing. There's a lot of brittle approaches here, Arbitrum doesn't give us a ton of assistance in terms of an API etc and the way they handle bridges unfolds across what are essentially three different approaches. I documented some of the process in the readme with links to example transactions.


packages/arbitrum/src/index.ts
```diff
@@ -0,0 +1,22 @@
+// This file is standard for more projects.
+// The main degree of nuance is in the
```
jimobrien, line 2: `is in the` 

I'm waiting in suspense!

packages/connext/package.json
```diff
@@ -15,9 +15,9 @@
     "bench": "vitest bench",
     "bench:ci": "CI=true vitest bench",
     "build": "pnpm run clean && pnpm run build:cjs && pnpm run build:esm && pnpm run build:types",
-    "build:cjs": "tsc --project tsconfig.build.json --module commonjs --outDir ./dist/cjs --removeComments --verbatimModuleSyntax false && echo > ./dist/cjs/package.json '{\"type\":\"commonjs\"}'",
-    "build:esm": "tsc --project tsconfig.build.json --module es2015 --outDir ./dist/esm && echo > ./dist/esm/package.json '{\"type\":\"module\",\"sideEffects\":false}'",
-    "build:types": "tsc --project tsconfig.build.json --module esnext --declarationDir ./dist/types --emitDeclarationOnly --declaration --declarationMap",
+    "build:cjs": "tsc --project tsconfig.build.json --module commonjs --moduleResolution node --outDir ./dist/cjs --removeComments --verbatimModuleSyntax false && echo > ./dist/cjs/package.json '{\"type\":\"commonjs\"}'",
```
jimobrien, line 18: reminder that we can remove these once #23 is merged

packages/arbitrum/src/chain-ids.ts
```diff
@@ -0,0 +1,5 @@
+// The arbitrum Native Bridge only supports three chains, Arbitrum, Arbitrum Nova, and Ethereum
+export const CHAIN_ID_ARRAY = [1, 42161, 42170]
```
jimobrien, line 2: any reason not to just use the named variables in the list? 

`export const CHAIN_ID_ARRAY = [ETH_CHAIN_ID, ARB_ONE_CHAIN_ID, ARB_NOVA_CHAIN_ID]`

jimobrien (approved):
LGTM

packages/arbitrum/src/chain-ids.ts
Quazia: It should be - good catch

PR #15, opened 2023-08-24
Fix(lint): fix uni lint error

jimobrien:
image


Quazia:
Is it possible there's a discrepancy between local and global settings? I saw a bunch of trailing semicolons get added onto another PR (which I actually prefer to as needed personally) once I get some of these bridge plugins under my belt I'll start trying to put some effort towards devops/upkeep on this repo.

PR #17, opened 2023-08-24
Feat(optimism): Optimism bridge

Quazia:
Still need to add more tests. There was a lot more assistance from published Optimism repos for this one, and even fewer networks to deal with. One complexity - they use slightly different functions if there's a recipient. Not quite sure how to handle that yet I'm going to have to figure it out in the testing but something to watch for.


packages/optimism/src/Optimism.ts
```diff
@@ -0,0 +1,54 @@
+import { type BridgeActionParams, compressJson } from '@rabbitholegg/questdk'
+import {
+  l1StandardBridgeABI,
+  l2StandardBridgeABI,
+  addresses,
+} from '@eth-optimism/contracts-ts'
+import { ETH_CHAIN_ID, CHAIN_ID_ARRAY } from './chain-ids.js'
+import { ETH_TOKEN_ADDRESS } from './token-addresses.js'
+// If you're implementing swap or mint, simply duplicate this function and change the name
```
jimobrien, line 9: can remove the template comments

packages/optimism/src/Optimism.ts
```diff
@@ -0,0 +1,54 @@
...
+      $abi: l2StandardBridgeABI,
+      _l2Token: tokenAddress,
+      _amount: amount,
+    }, // The input object is where we'll put the ABI and the parameters
+  })
+}
+
+export const getSupportedTokenAddresses = async (_chainId: number) => {
+  // Given a specific chain we would expect this function to return a list of supported token addresses
+  return []
```
jimobrien, line 49: this should return the tokens listed in packages/optimism/src/token-addresses.ts

Quazia: Should be resolved

packages/optimism/src/Optimism.ts
Quazia: Good catch thanks, I'll keep an eye out for loose comments in the other branches too.

jimobrien (approved):

PR #19, opened 2023-08-30
Feat(polygon): Polygon bridge

Quazia:
Biggest issue here is that the L1 ERC20 function signature doesn't have an amount. Polygon in general handles stuff super weird and I'm not very happy with this integration. They also have this concept around token predicates that is basically an alternative bridging flow but it's not supported through the UI.

Edit: Got the amount issue resolved on L1 ERC20. Not going to bother with the predicates right now as best I can tell they arn't used through the front-end anyhow.


packages/polygon/package.json
```diff
@@ -0,0 +1,52 @@
...
+    "ts-node": "^10.9.1",
+    "tsconfig": "workspace:*",
+    "typescript": "^5.1.6",
+    "vitest": "^0.33.0"
+  },
+  "dependencies": {
+    "@maticnetwork/pos-portal": "^1.5.2",
+    "@rabbitholegg/questdk": "1.0.1-alpha.8",
+    "viem": "^1.2.15"
+  }
```
JamieLottering, line 51: We should probably keep dependencies pinned for any plugins / sdks that other users will consume

packages/polygon/src/Polygon.test.ts
```diff
@@ -0,0 +1,139 @@
+import { GreaterThanOrEqual, apply } from '@rabbitholegg/questdk/filter'
+import { describe, expect, test } from 'vitest'
+import { bridge } from './Polygon.js'
+import { POLYGON_CHAIN_ID, ETH_CHAIN_ID, CHAIN_ID_ARRAY } from './chain-ids.js'
+import { MAINNET_BRIDGE } from './contract-addresses.js'
```
JamieLottering, line 5: Any reason for importing all these files with `.js` extensions included

packages/polygon/src/Polygon.ts
```diff
@@ -0,0 +1,58 @@
+import { type BridgeActionParams, compressJson } from '@rabbitholegg/questdk'
+import { POLYGON_BRIDGE_ABI_FUNCS } from './abi.js'
+import { POLYGON_CHAIN_ID, CHAIN_ID_ARRAY } from './chain-ids.js'
```
JamieLottering, line 3: Same comment here, can you not just do:
```
import { POLYGON_CHAIN_ID, CHAIN_ID_ARRAY } from './chain-ids'
```

packages/polygon/src/Polygon.ts
```diff
@@ -0,0 +1,58 @@
...
+    to: contractAddress || MAINNET_BRIDGE, // on Polgon the contract that handles withdrawals is the token contract
+    input: {
+      $abi: POLYGON_BRIDGE_ABI_FUNCS,
+      user: recipient,
+      rootToken: tokenAddress,
+    },
+  })
+}
+
+export const getSupportedTokenAddresses = async (_chainId: number) => {
```
JamieLottering, line 51: Does this need to be an `async` function?

packages/polygon/src/Polygon.ts
```diff
@@ -0,0 +1,58 @@
...
+  return PolygonTokens[_chainId]
+}
+
+export const getSupportedChainIds = async () => {
+  // This function should return a list of supported chainIds
+  return CHAIN_ID_ARRAY
+}
```
JamieLottering, line 58: Same here, I don't think this needs to be `async` unless I'm missing something

packages/polygon/src/supported-token-addresses.ts
```diff
@@ -0,0 +1,29 @@
+import { type Address } from 'viem'
+import { ETH_CHAIN_ID, POLYGON_CHAIN_ID } from './chain-ids.js'
+
+export const ETH_ADRESS_MAINNET = '0x0000000000000000000000000000000000000000'
```
JamieLottering, line 4: typo:
`ADRESS` -> `ADDRESS`

packages/polygon/src/supported-token-addresses.ts
```diff
@@ -0,0 +1,29 @@
...
+export const MATIC_ADDRESS_MAINNET =
+  '0x7D1AfA7B718fb893dB30A3aBc0Cfc608AaCfeBB0'
+export const MATIC_ADDRESS_POLYGON =
+  '0x0000000000000000000000000000000000001010'
+export const USDC_ADDRESS_MAINNET = '0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48'
+export const USDC_ADDRESS_POLYGON = '0x2791bca1f2de4661ed88a30c99a7a9449aa84174'
+export const WETH_ADDRESS_POLYGON = '0x7ceb23fd6bc0add59e62ac25578270cff1b9f619'
+// For now we're mainly supported the highest traffic Polygon tokens as listed here
+//https://github.com/maticnetwork/polygon-token-list
+export const PolygonTokens: Record<number, Address[]> = {
```
JamieLottering, line 14: Suggest to add `as const` for better IDE type hinting:

```
export const PolygonTokens: Record<number, Address[]> = {
  [ETH_CHAIN_ID]: [
    ETH_ADRESS_MAINNET, //ether
    MATIC_ADDRESS_MAINNET, //matic
    USDC_ADDRESS_MAINNET, //usdc
    '0xdac17f958d2ee523a2206206994597c13d831ec7', //usdt
    '0x6b175474e89094c44da98b954eedeac495271d0f', //dai
  ],
  [POLYGON_CHAIN_ID]: [
    WETH_ADDRESS_POLYGON, //WETH
    MATIC_ADDRESS_POLYGON, //matic
    USDC_ADDRESS_POLYGON, //usdc
    '0xc2132d05d31c914a87c6611c10748aeb04b58e8f', //usdt
    '0x8f3cf7ad23cd3cadbd9735aff958023239c6a063', //dai
  ],
} as const
```

packages/polygon/package.json
Quazia: Agreed - I'll capture this as an issue

packages/polygon/src/Polygon.test.ts
Quazia: Compiler was complaining when I didn't have them on there, here's the error:
```
Relative import paths need explicit file extensions in EcmaScript imports when '--moduleResolution' is 'node16' or 'nodenext'. Did you mean './types.js'
```

Lmk if there's a smarter way to be fixing this, I agree it looks weird _especially_ since they're `.ts` files

packages/polygon/src/Polygon.ts
Quazia: The type definition specifies `async`, but this function does not need to be.
```
export interface IActionPlugin {
  pluginId: string
  getSupportedChainIds: () => Promise<number[]>
  getSupportedTokenAddresses: (chainId: number) => Promise<Address[]>
  bridge: (
    params: BridgeActionParams,
  ) => Promise<TransactionFilter> | Promise<PluginActionNotImplementedError>
  swap: (
    params: SwapActionParams,
  ) => Promise<TransactionFilter> | Promise<PluginActionNotImplementedError>
  mint: (
    params: MintActionParams,
  ) => Promise<TransactionFilter> | Promise<PluginActionNotImplementedError>
}
```

What do you think, better to just drop it in the definition still? I'm ambivalent, as long as it's consistent how we handle it.

packages/polygon/src/supported-token-addresses.ts
Quazia: Thanks!

packages/polygon/src/supported-token-addresses.ts
Quazia: Huh, okay will do. Thanks!

packages/polygon/src/Polygon.ts
JamieLottering: Ok cool, lets keep it consistent then! `async` it is

packages/polygon/src/supported-token-addresses.ts
JamieLottering: Personal preference on my part, really doesn't matter in the end. I just love being able to hover over types and see exactly what I'm getting

packages/polygon/src/Polygon.test.ts
jimobrien: ESM modules require you to specify the file extension

packages/polygon/src/supported-token-addresses.ts
Quazia: Done-zo
When is this necessary? Is it specifically when I'm keying off of variables or what's like the reason I need to tack on the `as const` for better type hinting?

jimobrien (approved):

PR #20, opened 2023-09-01
Feat(hop): Hop bridge

Quazia:
Adds hop support for bridge action


jimobrien (approved):

packages/hop/src/Hop.ts
Quazia, line 1: @JamieLottering Could you look at how I'm handling the typing in this file? Just want to make sure there's not a better way to be handling this without having to force so many casts.

PR #22, opened 2023-09-08
Feat(stargate): Stargate bridge

Quazia:
Some of the data in `CHAIN_AND_POOL_TO_TOKEN_ADDRESS` in the `contract-addresses.ts` file are the pool token instead of the actual token. Going to have to fix the data in that lookup before we merge this in.

Fixed the  `CHAIN_AND_POOL_TO_TOKEN_ADDRESS` for `USDC` `USDT` and `WETH`. Should be G2G.


packages/stargate/src/Stargate.ts
```diff
@@ -0,0 +1,67 @@
...
+    tokenAddress,
+    amount,
+    recipient,
+  } = bridge
+  const layerOneChainId = LAYER_ZERO_TO_LAYER_ONE_CHAIN_ID[sourceChainId]
+  const sourcePool = tokenAddress
+    ? CHAIN_AND_POOL_TO_TOKEN_ADDRESS[sourceChainId][tokenAddress]
+    : 0
+
+  if (sourcePool === 13) {
```
jimobrien, line 30: is this `const ETH_POOL = 13` accurate?

Quazia: Yep, I'll create a const and compare against that vs a magic number for increased clarity.

jimobrien (approved):

PR #32, opened 2023-09-22
Fix(stargate): supply layer one chain ids

Quazia:
The bridge function previously expected layer zero values to be fed in, this resolves that issue.


packages/stargate/src/contract-addresses.ts
```diff
@@ -176,3 +185,43 @@ export const CHAIN_AND_POOL_TO_TOKEN_ADDRESS: {
     '0x224d8fd7ab6ad4c6eb4611ce56ef35dec2277f03': ETH_POOL_ID,
   },
 }
+
+export const CHAIN_ONE_AND_POOL_TO_TOKEN_ADDRESS: {
```
jonathandiep, line 189: [very-nit] Not sure if `CHAIN_ONE` is best description for these. Wondering if we should use something like `NATIVE_CHAIN_AND_POOL_TO_TOKEN_ADDRESS`?

jonathandiep (approved):

PR #41, opened 2023-09-27
Feat(tally): add support for tally delegate

Quazia:
*~*~*WIP*~*~*

Based off of https://github.com/rabbitholegg/questdk-plugins/pull/40 which should be brought upstream first and is blocked by [changes in questdk](https://github.com/rabbitholegg/questdk/pull/20)

### DIFF
--- a/packages/registry/src/index.test.ts
+++ b/packages/registry/src/index.test.ts
@@ -34,6 +34,7 @@ describe('executePlugin', () => {
       bridge: vi.fn(),
       swap: vi.fn(),
       mint: vi.fn(),
+      delegate: vi.fn(),
     } as unknown as IActionPlugin
 
     await executePlugin(mockPlugin, ActionType.Bridge, mockParams)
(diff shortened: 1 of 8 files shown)


packages/tally/src/token-addresses.ts
```diff
@@ -0,0 +1,10 @@
+import { type Address } from 'viem'
+import { ARB_ONE_CHAIN_ID } from './chain-ids.js'
+
+// For now we're mainly supported the highest traffic Arbitrum tokens
+// Support for USDC seems like it might be non-trivial
+// https://docs.arbitrum.io/bridge-tokens/concepts/usdc-concept
+// [DAI]
```
jonathandiep, line 7: [nit] I think this comment was carried over from Arbitrum bridge?

Quazia: 100% good catch thanks

PR #43, opened 2023-10-01
feat(basepaint): add support for basepaint mint

Mmackz:
Pretty trivial implementation for this one. 

- [Contract Address](https://basescan.org/address/0xba5e05cb26b78eda3a2f8e3b3814726305dcac83)
- [Sample Tx](https://basescan.org/tx/0x82b093ba5ca167e86ce3c7620d9c705b7f0416c377b3975693aa070dad63b082)

inputdata

The 2 values in the input data `day` and `count` map to the mint action params `tokenId` and `quantity` respectively, which made for a simple implementation.

I have confirmed that the `day` value in the input data does match with the token id, ever since day 1.


packages/basepaint/src/Basepaint.test.ts
```diff
@@ -0,0 +1,58 @@
...
+          $abi: MINT_ABI,
+          day: 69,
+          count: {
+            $gte: '1',
+          },
+        },
+      })
+    })
+
+    test('should pass filter with valid transactions', async () => {
```
Quazia, line 28: Can we move the transaction object to the constants file and add a second transaction that it fails for?

Quazia (approved):
LGTM! Two small nits and let's get this in. SUPER appreciate you for doing this!

packages/basepaint/src/Basepaint.test.ts
Mmackz: Done! I added additional tests to validate that the filter correctly rejects transactions with an incorrect tokenId or quantity.

PR #49, opened 2023-10-07
Fix(stargate): Add more tokens

jonathandiep:
Adds ETH/WETH to supported networks and includes USDC for Optimism

### DIFF
--- a/packages/stargate/src/Stargate.ts
+++ b/packages/stargate/src/Stargate.ts
@@ -7,6 +7,7 @@ import {
   ETH_CHAIN_ID,
   LAYER_ONE_TO_LAYER_ZERO_CHAIN_ID,
   POLYGON_CHAIN_ID,
+  OPTIMISM_CHAIN_ID,
 } from './chain-ids.js'
 import {
   NATIVE_CHAIN_AND_POOL_TO_TOKEN_ADDRESS,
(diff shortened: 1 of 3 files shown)


packages/stargate/src/Stargate.ts
```diff
@@ -69,11 +70,25 @@ export const getSupportedTokenAddresses = async (
...
   if (_chainId === POLYGON_CHAIN_ID)
-    return ['0x2791Bca1f2de4661ED88A30C99A7a9449Aa84174']
+    return [
+      '0x2791Bca1f2de4661ED88A30C99A7a9449Aa84174',
+      '0x7ceb23fd6bc0add59e62ac25578270cff1b9f619',
+    ] // [UDSC, WETH]
   if (_chainId === ARBITRUM_CHAIN_ID)
-    return ['0xaf88d065e77c8cC2239327C5EDb3A432268e5831']
+    return [
+      '0xaf88d065e77c8cC2239327C5EDb3A432268e5831',
```
Quazia, line 79: https://docs.arbitrum.io/bridge-tokens/concepts/usdc-concept

Do we want USDC or USDC.e?

Quazia (approved):
LGTM - assuming we want USDC.e vs USDC in both cases

packages/stargate/src/Stargate.ts
jonathandiep: I think Stargate only accepts USDC (at least I don't see an entry for USDC.e)

PR #55, opened 2023-10-10
Add registry contracts

Quazia:
This adds an endpoint in the registry to handle the indexed contracts use case. For now it just returns the entrypoint contract for account abstraction, but in the future we'll want to support multi-call and other utility contracts that allow for delegated call flows.


jonathandiep (approved):

packages/registry/src/contract-addresses.ts
```diff
@@ -0,0 +1 @@
+export const ENTRYPOINT = '0x5ff137d4b0fdcd49dca30c7cf57e578a026d2789'
```
jonathandiep, line 1: Can we assume that this address is deployed on the same address for all evm equivalent chains?

If not then looks like we can manage using `chainId` later

Quazia: It is the same on all chains

PR #56, opened 2023-10-12
Feat(camelot): add swap support

Quazia:


packages/camelot/src/Camelot.test.ts
```diff
@@ -0,0 +1,54 @@
...
+    })
+    test('should pass filter with valid ETH transactions', async () => {
+      const transaction = SWAP_ETH
+      const filter = await swap({
+        chainId: ARBITRUM_CHAIN_ID,
+        contractAddress: CAMELOT_ROUTER,
+        tokenOut: '0xBfbCFe8873fE28Dfa25f1099282b088D52bbAD9C',
+        amountOut: GreaterThanOrEqual(parseEther('0.0005')),
+      })
+      console.log(filter)
```
jonathandiep, line 39: Can remove this log

packages/camelot/src/abi.ts
```diff
@@ -0,0 +1,247 @@
...
+      { internalType: 'uint256', name: 'amountAMin', type: 'uint256' },
+      { internalType: 'uint256', name: 'amountBMin', type: 'uint256' },
+      { internalType: 'address', name: 'to', type: 'address' },
+      { internalType: 'uint256', name: 'deadline', type: 'uint256' },
+      { internalType: 'bool', name: 'approveMax', type: 'bool' },
+      { internalType: 'uint8', name: 'v', type: 'uint8' },
+      { internalType: 'bytes32', name: 'r', type: 'bytes32' },
+      { internalType: 'bytes32', name: 's', type: 'bytes32' },
+    ],
+    name: 'removeLiquidityWithPermit',
```
jonathandiep, line 197: Dumb q: If we include the other functions inside the abi, is it possible that they could trigger a successful action? Like if the params match

e.g. Can this `removeLiquidity` fx trigger a completion for a `swap` action if it had matching params?

jonathandiep (approved):

packages/camelot/src/abi.ts
Quazia: Yeah that's possible - I just narrowed the included ABIs that's easier than seeing if what you're describing could happen lol

packages/camelot/src/Camelot.test.ts
Quazia: I removed a few other hanging ones - I'm so bad at removing logs I'd make a terrible lumberjack

Issue #63, opened 2023-10-14
bug(stargate): Incorrect Filtering Due to Undefined `sourcePool` when Using wETH


Mmackz:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Package Version

2.0.0-alpha.18

**Affected Plugin** - Stargate Plugin

#### Description:

The current implementation uses wETH as the default token for bridging operations. However, many of the supported chains do not have support for wETH. As a result, the `sourcePool` can become `undefined` when applying a filter with wETH as the token address.

This will cause unexpected outcomes, such as having unspecified tokens incorrectly passing the filter.

#### Proposed Solution:
1. **Use ETH Instead of wETH:** Given its widespread availability across most supported chains.
2. **Omit Unsupportive Chains from UI:** Omit the option for either ETH or wETH on chains that don't support them to ensure they don't appear in the UI.
3. **Fallback Mechanism for `sourcePool`:** Implement a general fallback mechanism to handle scenarios where `sourcePool` is `undefined`.

PR #64, opened 2023-10-16
fix(stargate): Replace wETH with ETH and Improve Error Handling in Stargate Plugin

Mmackz:
Fixes: #63

**Changes Made:**

- Replaced wETH with ETH in the `getSupportedTokenAddresses` function.
- Reject unsupported or undefined `tokenAddress`: An error will be thrown if a `tokenAddress` is unsupported or undefined.
- Expanded test coverage: Tests have been added to validate that each token address returned from `getSupportedTokenAddresses` produces a valid filter.
- Replaced unsupported token USDC on arbitrum with USDC.e
- Removed Polygon due to it not supporting ETH transfers

**Reason for Changes:**
The original implementation used wETH as the default token for bridging operations, leading to potential filtering issues due to wETH being unsupported by stargate on several chains. Additional tests where added to detect unsupported tokens returned by the `getSupportedTokenAddresses` function. These changes aim to rectify the previous issues and ensure more accurate filtering and test coverage.


Mmackz:
Added a check to verify that the argument `destinationChainId` produces a valid LayerZero Chain Id. If not, the function will throw an error.

packages/stargate/src/Stargate.ts
Mmackz, line 1: Added a check to verify that the argument destinationChainId produces a valid LayerZero Chain Id. If not, the function will throw an error.

Quazia:
I merged this in but we have some changes in the architecture to allow for the `to` field to be an array which changes the potential solutions here. We also ideally don't want to throw errors in these plugins, better to return null or fail gracefully and return a message. The section of code upstream doesn't catch the error and the error type could be anything with a plug-in theoretically.

PR #83, opened 2023-10-24
fix: remove private in order to publish

jonathandiep:
Unblocks build issue when trying to pull latest version


Quazia (approved):
LGTM - We gotta test this with the indexer/terminal still

PR #85, opened 2023-10-25
feat(symbiosis): add support for symbiosis bridge

mmackz:
## Integration of `symbiosis` plugin for QuestDK

This PR introduces the `symbiosis` plugin integration for QuestDK. 

#### About The Project
[**Symbiosis Bridge**](https://app.symbiosis.finance/), is a cross-chain bridge protocol that is compatible with many networks. 

### Implementation Details

This plugin will support bridging of most stable coins and popular tokens on each chain. The filter takes into account the input token, the source chain and destination chain to determine if a transaction is valid. The output token can be chosen freely by the user and should not affect the validity of the transaction.


packages/symbiosis/package.json
```diff
@@ -0,0 +1,53 @@
+{
+  "name": "@rabbitholegg/questdk-plugin-symbiosis",
+  "private": true,
```
jonathandiep, line 3: Lets remove this `private` setting so that this plugin can be published

packages/symbiosis/src/Symbiosis.test.ts
```diff
@@ -0,0 +1,110 @@
...
+                },
+              },
+            },
+          },
+        },
+      })
+    })
+  })
+  describe('should pass filter with valid transactions', () => {
+    PASSING_TEST_TRANSACTIONS.forEach((testTransaction) => {
```
Quazia, line 54: I love this pattern - NICE
Gotta convert the other plugins to use this this is slick

packages/symbiosis/src/Symbiosis.ts
```diff
@@ -0,0 +1,59 @@
...
+
+  return compressJson({
+    chainId: sourceChainId,
+    to: bridgeContract,
+    input: {
+      $abi: metaRouteABI,
+      _metarouteTransaction: {
+        approvedTokens: tokenAddress ? [tokenAddress] : undefined, // if tokenAddress is undefined, any input token will pass filter
+        amount: amount,
+        otherSideCalldata: {
```
Quazia, line 34: Right here you should be able to do 
```
otherSideCalldata: {
    $abiAbstract: metaBurnABI,
    _metaBurnTransaction: {
        chainID: destinationChainId,
        chain2address: recipient,
    }
}
```

And that should work - I wouldn't bother cause this looks like it's working fine and it's more precise but just to see how you'd use that.

Quazia (approved):
LGTM - really good work I really love the new testing pattern you're using here.

packages/symbiosis/src/Symbiosis.ts
mmackz: I like this solution much better, and it allows us to remove the abi for the metaSynthesize function.

Im going to go ahead and implement this.

mmackz:
Private setting is removed 👍

Moved the pr out of the draft state.

PR #90, opened 2023-10-31
Feat(rabbithole): initial quest plugin

Quazia:


packages/rabbithole/src/contract-addresses.ts
```diff
@@ -0,0 +1,34 @@
+export const RABBITHOLE_QUEST_FACTORY =
```
mmackz, line 1: Keeping the exports below the imports for readability purposes.

packages/rabbithole/src/test-transactions.ts
```diff
@@ -0,0 +1,77 @@
...
+  v: 1n,
+  value: 0n,
+  typeHex: '0x2' as Hash,
+  yParity: 0,
+} as TransactionEIP1559
+
+export const controlTransaction: TestTransaction = {
+  transaction: CREATE_AND_QUEUE_QUEST_TX,
+  rewardToken: '0x4200000000000000000000000000000000000042',
+  rewardAmount: 1500000000000000000,
```
mmackz, line 42: It doesnt appear to be causing any issues, but this number is too large for a standard integer.

mmackz (approved):
Looks Good!!

mmackz:
added a few tests to make sure undefined parameters don't cause any issues.

packages/rabbithole/src/test-transactions.ts
Quazia: Converted that and a few others to `BigInt` just to be safe 😊 Good catch!

PR #100, opened 2023-11-13
feat(paraswap): enhance transaction support and improve token lists

mmackz:
#### Description

This pull request introduces several enhancements to the ParaSwap swap plugin, addressing previously identified gaps in transaction monitoring and expanding its capabilities to improve user experience. 

#### Implemented Enhancements

1. **Improved Megaswap and BalancerV2 Swaps Method Detection:**
   - **Changes Made:** Updated the plugin to accurately detect and handle transactions involving Megaswap and BalancerV2 methods.
   - **Outcome:** Enhanced reliability in transaction monitoring for these methods.

2. **Added Support for Raw ETH Transactions:**
   - **Implementation Details:** Integrated functionality to handle raw ETH transactions.
   - **Impact:** This addition ensures complete transaction monitoring coverage, including ETH-based transactions.

#### Additional Updates

- **Removal of AAVE aTokens:** AAVE aTokens have been removed from the default token list.
- **Curated Token List for Ethereum:** A list of popular tokens for Ethereum has been curated. This is due to the sheer number of addresses returned by the api. (5000+)
- **Expanded Test Coverage:** Added tests for additional swap routes, and that a valid token list is returned for each supported chain.


packages/paraswap/src/Paraswap.ts
```diff
@@ -22,96 +23,66 @@ import {
...
-    amountIn,
-    amountOut,
-    recipient,
-  } = swap
+  const { chainId, contractAddress, tokenIn, tokenOut, amountIn, amountOut } =
+    swap
   const { getAugustusSwapper } = constructGetSpender({ chainId, fetcher })
   const to = contractAddress || (await getAugustusSwapper())
-  // We always want to return a compressed JSON object which we'll transform into a TransactionFilter
+  const ethUsedIn = tokenIn === Tokens.ETH
```
Quazia, line 33: Is there ever a case where if `tokenIn === Tokens.ETH` we don't use `ETH_ADDRESS`? Cause you could just do like...

```
const tokenInUsed = tokenIn === Tokens.ETH : ETH_ADDRESS ? tokenIn
```

And just always use that value to avoid needed the extra logic in each ABI param section.

In cases like ` assets: buildPathQuery(tokenIn, tokenOut),` on line 84 we could just still use the `tokenIn` (assuming that isn't supposed to use the `ETH_ADDRESS`

Quazia (approved):
This is good for now but we're going to need to add the [entrypoint ](https://github.com/eth-infinitism/account-abstraction/blob/develop/contracts/core/EntryPoint.sol) address to the `to` and make it an `abiAbstract` again at some point

packages/paraswap/src/Paraswap.ts
mmackz: I have made the recommended adjustments. Please take a quick look when you get a chance. 🙏

Quazia:
Looks great! Thank you

PR #106, opened 2023-11-17
Synapse SDK Plugin (Draft)

Defi-Moses:
This is a first attempt at creating the questsdk for the Synapse plugin. There is support for the original synapse bridge contracts as well as an outline for supporting the SynapseCCTP contracts. This version still needs robust testing support.


packages/synapse/src/Token.ts
```diff
@@ -0,0 +1,185 @@
+
+
+export class Token {
```
Quazia, line 3: Same thing here as the `tokens.ts` folder.

packages/synapse/src/Synapse.ts
```diff
@@ -0,0 +1,83 @@
...
+    destinationChainId,
+    contractAddress,
+    tokenAddress,
+    amount,
+    recipient,
+  } = bridge
+
+  const destinationDomain = chainDomainToID(destinationChainId)
+
+  if (contractAddress && Object.values(cctpMapping).includes(contractAddress)) {
```
Quazia, line 33: Okay so basically the domain is only defined on some chains and this covers that?

Quazia (commented):
This looks like you're definitely going in the right direction! There might be some peculiarities as you enter testing but I don't forsee any major hiccups.
Two things to flag:
1. if at all possible for the `tokens.ts` and `Tokens.ts` file if those values are available via an API, or a published repo the plugin will be a lot more stable relying on that source.
2. Make sure that nothing breaks if any of the optional values on ` BridgeActionParams` are passed in as undefined.

Thanks so much for taking the time to draft a plugin for our system!

packages/synapse/src/Synapse.ts
Defi-Moses: There are two different bridge contracts -- this is a case to handle when a transaction is routed through the "SynapseCCTP bridge"

packages/synapse/src/Synapse.ts
```diff
@@ -0,0 +1,83 @@
...
+  return supportedTokens.map(token => token.addresses[chainId]) as Address[];
+}
+
+
+export const getSupportedChainIds = async (): Promise<number[]> => {
+  return CHAIN_ID_ARRAY
+}
+
+//Need to edit based on CCTP domains
+export const chainDomainToID = (id: number): number => {
```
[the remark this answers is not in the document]
Defi-Moses: Only transactions routed through the "SynapseCCTP bridge" will have the parameter "chainDomain" thus we need to convert that into a chainID for the rest of the application

packages/synapse/src/tokens.ts
```diff
@@ -0,0 +1,1117 @@
+import { zeroAddress } from 'viem'
```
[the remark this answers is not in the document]
Defi-Moses: Thanks for the heads up -- for now will update manually but wokring on a clean solution to export it from the source.

PR #108, opened 2023-11-18
feat(arbitrum): add support for native arbitrum bridge

mmackz:
#### Description

This pull request introduces the `arbitrum` plugin integration for QuestDK.

#### Implementation Details
- **Testing:** Testing covers both Deposits and Withdrawals of ETH and Tokens, and makes sure that undefined parameters will not cause any issues.
- **Chain Compatibility:** Ethereum must be either the sourceChain or destinationChain. Bridging directly between Arbitrum Nova and Arbitrum One is not possible.
- **Token Selection Criteria:** Only tokens compatible with all three chains have been added. Notably, USDT was excluded due to issues with bridging to Arbitrum Nova.

#### Technical Notes

- **Challenges with overloaded function:** I attempted to combine the logic in the bridge filter function, but ran into issues due to the overloaded function `outboundTransfer`. It is a little messy, but it works.

- **Setting `tokenAddress` to "any":** A lot of logic, such as determining the bridge contract address, and determining the amount (no context for `decimals`) relies on the `tokenAddress` being explicitly set. Because of this, when setting `tokenAddress` to "any", both the `destinationChain` and `amount` will also be "any". This will lead to some false-positives, but should not negatively affect user experience.


Quazia (approved):
This LGTM but I think it'll be worth re-visiting this plugin after we have the ability to do any <-> any network bridging. 

I think we can basically make it even more general so for this:
```
    const contracts =
      sourceChainId === ETH_CHAIN_ID
        ? [
            MAINNET_TO_ARB_NOVA_GATEWAY,
            MAINNET_TO_ARB_ONE_GATEWAY,
            ARB_ONE_DELAYED_INBOX,
            ARB_NOVA_DELAYED_INBOX,
          ]
        : [
            ARB_ONE_TO_MAINNET_GATEWAY,
            ARB_NOVA_TO_MAINNET_GATEWAY,
            UNIVERSAL_ARBSYS_PRECOMPILE,
          ]
          ```
          
          If the chain ID is undefined we'd just need to be checking the superset of these two address arrays.

PR #112, opened 2023-11-27
feat(okutrade): add support for Oku Trade plugin

mmackz:
## Integration of `okutrade` swap plugin for QuestDK

This PR introduces the integration of the `okutrade` swap plugin for QuestDK.

### About The Project

Oku Trade is a trader-centric front-end for uniswaps universal router contracts. It aims to brig enhanced analytics and charts to uniswap traders through there enhanced front end.

### Implementation Details

The logic for this plugin is mostly the same as the uniswap plugin, as the both target the same functions on the universal router contract. Oku Trade uses the pre-migration version of these contracts, so the contract address will be different from uniswaps. The plugin has been thoroughly tested to ensure valid transactions originating from the oku trade front end will pass the filter. Additional testing is done to make sure that undefined parameters do not cause any unintended false negatives. 

### Current Limitations

- None. Everything should work as expected.


Quazia (approved):
LGTM - you basically just copied over most of this from the Uniswap then tweaked the `chainToContract` addresses? Anything else major that needed revision?

PR #114, opened 2023-11-30
feat(balancer): add support for balancer swap

mmackz:
## Integration of `balancer` Plugin for QuestDK

This PR introduces the integration of the `balancer` plugin for QuestDK.

### About The Project
Balancer is a popular ethereum based dex which is deployed on several L2 networks. It functions primarily as a decentralized exchange using an automated market maker (AMM) algorithm to facilitate efficient swapping of tokens.

### Implementation Details

This plugin supports several popular tokens on each supported chain. All tokens have been manually tested in the UI to ensure there is sufficient liquidity for trading. It has been thoroughly tested to ensure functionality across both `swap` and `batchSwap` methods. 

### Current Limitations

- for swaps using the basic `swap` method, determining the `amountOut` is not possible, due to it not being included in the input call data (It is returned in the output). This will lead to some false-positives, but should not negatively affect user experience.


packages/balancer/src/Balancer.ts
```diff
@@ -0,0 +1,63 @@
...
+    tokenIn,
+    tokenOut,
+    amountIn,
+    amountOut,
+    recipient,
+  } = swap
+
+  return compressJson({
+    chainId: chainId,
+    to: contractAddress ?? VAULT_CONTRACT,
```
Quazia, line 27: Can you do me a favor and add the `0x5ff137d4b0fdcd49dca30c7cf57e578a026d2789` entry point contract as an array for the `$to` here - patch wallet wants to use this plug-in to test account abstraction with plugins.

packages/balancer/src/constants.ts
```diff
@@ -0,0 +1,11 @@
+import { Chains } from './utils'
+
+export const VAULT_CONTRACT = '0xBA12222222228d8Ba445958a75a0704d566BF2C8'
```
Quazia, line 3: Is this the right case?

packages/balancer/src/constants.ts
```diff
@@ -0,0 +1,11 @@
+import { Chains } from './utils'
```
Quazia, line 1: This is going to make it super easy to upgrade to the internal utils package - nice

packages/balancer/src/utils.ts
```diff
@@ -0,0 +1,127 @@
...
+  if (tokenOut) {
+    conditions.push({ $last: getAddress(tokenOut) })
+  }
+
+  return {
+    $and: conditions,
+  }
+}
+
+export const buildAmountQuery = (
```
Quazia, line 83: This is slick

Quazia (approved):
LGTM - one small request but nothing I see that's incorrect.
That `buildAmountQuery` is cool

packages/balancer/src/constants.ts
mmackz: I'll change it to lowercase since we will be using the $or array.

PR #117, opened 2023-12-04
Feat(zora): add mint plugin

Quazia:
There's two main contracts, one for 721 and one for 1155. This supports both.

The filter requires that a contract address is found since every mint goes through a different contract. There might be a way to get the full list but I don't think we want to support minting on any contract at this time.

It looks like there's collections with legacy contracts, I tried to support the ones I could find but it's possible that this plugin will break for collections launched too long into the past. Should work for all recent collections and all going forward.


packages/zora/src/Zora.test.ts
```diff
@@ -0,0 +1,34 @@
...
+import { describe, expect, test } from 'vitest'
+import { failingTestCases, passingTestCases } from './test-setup'
+import { mint } from './Zora'
+
+describe('Given the zora plugin', () => {
+  describe('When handling the mint', () => {
+
+    test('should return a valid action filter', () => {
+      
+    })
```
jonathandiep, line 11: Still need this?

packages/zora/src/index.ts
```diff
@@ -0,0 +1,23 @@
...
+} from '@rabbitholegg/questdk'
+
+import {
+  mint,
+  getSupportedChainIds,
+  getSupportedTokenAddresses,
+} from './Zora.js'
+
+// Replace *project* with the name of the project
+export const {Zora}: IActionPlugin = {
```
mmackz, line 16: This should not be destructured

packages/zora/src/test-setup.ts
```diff
@@ -0,0 +1,32 @@
...
+    chainId: 1,
+  }),
+  createTestCase(BASIC_PURCHASE, 'when contractAddress is incorrect', {
+    contractAddress: '0x4f330940159fB3368F5b06b34212C0cDB4e2C032',
+  }),
+  createTestCase(BASIC_PURCHASE, 'when recipient is incorrect', {
+    recipient: '0xd31143Ca8503b25DdE780dc1B92E9aA61D0E326d',
+  }),
+  createTestCase(MINT_WITH_REWARDS_1155, 'when tokenId is incorrect', {
+    tokenId: '1',
```
mmackz, line 26: tokenId expects a number here.

packages/synapse/package.json
jonathandiep, line 1: Not sure if you meant to include this file?

packages/zora/src/test-transactions.ts
```diff
@@ -0,0 +1,60 @@
...
+    input:
+      '0x9dbb844d00000000000000000000000004e2516a2c207e84a1839755675dfd8ef6302f0a0000000000000000000000000000000000000000000000000000000000000019000000000000000000000000000000000000000000000000000000000000000100000000000000000000000000000000000000000000000000000000000000a000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000020000000000000000000000000ed12545c21a85bab43b25128f49f81446776f33e',
+    to: '0x4c0c2DD31d2661E8BCeC60a42e803dCc6F81Baad',
+    value: '22977000000000000',
+  },
+  params: {
+    chainId: Chains.ZORA,
+    contractAddress: '0x4c0c2DD31d2661E8BCeC60a42e803dCc6F81Baad'.toLowerCase() as Address,
+    amount: '1',
+    tokenId: '25',
```
mmackz, line 58: tokenId should be a number

packages/zora/src/Zora.test.ts
Quazia: I should at a filter test - yeah good callout

packages/synapse/package.json
Quazia: Deff not I should probably remove it lol

packages/zora/src/Zora.ts
```diff
@@ -0,0 +1,46 @@
...
+import { type Address } from 'viem'
+import { CHAIN_ID_ARRAY } from './chain-ids'
+import { ZORA_MINTER_ABI } from './abi'
+
+export const mint = async (
+  mint: MintActionParams,
+): Promise<TransactionFilter> => {
+  const { chainId, contractAddress, tokenId, amount, recipient } = mint
+
+  let andArray = []
```
mmackz, line 15: The linter wants this to be a const, since it does not get reassigned. Not really a big deal though as it still passes the tests.

PR #118, opened 2023-12-04
Feat(synapse): add bridge

Quazia:


packages/synapse/src/Synapse.test.ts
```diff
@@ -0,0 +1,235 @@
...
+        })
+      }
+
+    describe('When applying the filter', () => {
+      test('should pass filter with valid L1 ETH tx', async () => {
+        const transaction = DEPOSIT_ETH
+        const filter = await bridge({
+          sourceChainId: ETH_CHAIN_ID,
+          destinationChainId: ARBITRUM_CHAIN_ID,
+          tokenAddress: '0xEeeeeEeeeEeEeeEeEeEeeEEEeeeeEeeeeeeeEEeE',
```
mmackz, line 175: For ETH to work properly on our front end, it needs to use the zero address.

You will need to convert it back to 0xEeee in the bridge function before returning the filter.

packages/synapse/src/Synapse.ts
```diff
@@ -0,0 +1,88 @@
...
+    tokenAddress,
+    amount,
+    recipient,
+  } = bridge
+
+  const inputObject = {
+    amount: amount,
+    chainId: destinationChainId,
+    token: tokenAddress,
+  }
```
mmackz, line 32: Something like this:
```suggestion
    token:
      tokenAddress === zeroAddress
        ? '0xEeeeeEeeeEeEeeEeEeEeeEEEeeeeEeeeeeeeEEeE'
        : tokenAddress,
```

packages/synapse/src/Synapse.test.ts
```diff
@@ -60,176 +59,192 @@ describe('When given the Synapse plugin', () => {
         },
       })
     })
-    test(
```
mmackz: These tests were not being applied properly. The async function needs to be passed into the test function as an argument.

Also the filter needed to be updated to use the $or pattern.

I already went ahead and made the required changes.

mmackz:
Everything is good to go, I went ahead and made the necessary changes.

PR #119, opened 2023-12-04
fix(basepaint): use updated mint action params

mmackz:
Updates basepaint package to use new mint action params


Quazia (approved):
LGTM - thanks for this

PR #120, opened 2023-12-07
feat(woofi): add support for WooFi swap plugin

mmackz:
## Integration of `woofi` swap action plugin for QuestDK

This PR introduces the integration of the `woofi` swap plugin for QuestDK.

### About The Project

WOOFi is a cross-chain DEX where anyone can swap, stake, and earn crypto and trade perpetual futures across every major blockchain.

### Implementation Details

The logic for the swap action on woofi is very straight forward and the inputs map really well to our actionparams. There is two kinds of swaps we see on the WooRouter contract, which are `swap` and `externalSwap`. Both use the same parameters.

### Current Limitations

- If the user sets a high slippage, amountOut will be lower than expected due to the way it is calculated (minAmountOut).


Quazia (approved):
Looks great as usual - if you could resolve the conflicts then just go ahead and merge that would be great.

PR #123, opened 2023-12-09
feat(zora): add support for batch mints through the universal minter contract

mmackz:
This pr adds support for NFTs which are minted through a batch minting process. 

Batch mints are routed through the universal minter contract. 

Here is the universal minter contract on optimism - https://optimistic.etherscan.io/address/0x97eb05b8db496b12244bccf17cf377d00a99b67a#code

Specifically, `mintBatchWithoutFees` is the function that was added.


packages/zora/package.json
```diff
@@ -44,6 +44,7 @@
   },
   "dependencies": {
     "@rabbitholegg/questdk": "2.0.0-alpha.28",
+    "@zoralabs/universal-minter": "^0.2.12",
```
Quazia, line 47: Let's pin this

packages/zora/src/Zora.ts
```diff
@@ -3,15 +3,25 @@ import {
...
   mint: MintActionParams,
 ): Promise<TransactionFilter> => {
   const { chainId, contractAddress, tokenId, amount, recipient } = mint
 
+  const universalMinter = zoraUniversalMinterAddress[
+    chainId as Chains
+  ].toLowerCase() as Address
+
+  if (!universalMinter) {
+    throw new Error(`no universal minter contract found for ${chainId}`)
```
Quazia, line 22: I think it's better to just fail-safe in this case. Can we do a ternary and only pass the array if the `universalMinter` is defined?

Quazia (approved):
LGTM just pin the package and tweak mint to fail-safe and we're good

PR #124, opened 2023-12-09
Feat(utils): add utils

Quazia:
Gives us a `utils` package to put shared default token lists, chain lists, and basic repeated helpers in.


jonathandiep (approved):
lgtm!

May need to include a changeset?

PR #125, opened 2023-12-09
fix(okutrade): convert V3 token addresses to lowercase

mmackz:
This PR fixes an issue where some token swaps were not being detected due to a casing issues.


packages/okutrade/src/utils.ts
jonathandiep, line 1: Do we also need to add `toLowerCase()` on the tokenIn/tokenOut for the `buildV2PathQuery` function? (lines 81 & 85)

jonathandiep (approved):
lgtm, but left a comment on if we may need to apply same fix elsewhere

packages/okutrade/src/utils.ts
mmackz: V2 expects a checksummed addess. It should be good.

PR #127, opened 2023-12-12
fix(uniswap): fix casing issue with tokens on V3 swaps

mmackz:
This PR fixes an issue where some token swaps were not being detected due to a casing issue.


Quazia (approved):
LGTM

PR #129, opened 2023-12-14
feat(vela): add support for Vela derivatives plugin

mmackz:
## Integration of `vela` derivative action plugin for QuestDK

This PR introduces the integration of the `vela` derivative plugin for QuestDK.

### About The Project

Vela Exchange is a decentralized trading platform with its foundations in the Arbitrum ecosystem. 

### Implementation Details

~~Because Vela is not a traditional DEX, we needed to make some compromises to make it fit into our swap action. Because it uses USDC pairs, there is no way to swap between two supported tokens. Because of this, we only make sure that the input token was part of the trade (either short or long).~~ Will now use the new `OptionsActionParams` instead of shoe-horning it into our existing `SwapActionParams`

### Current Limitations

~~- TokenOut will always be set to "any" regardless of which token is selected.~~
~~- AmountOut will always be set to "any" regardless of which amount is set.~~

- Everything should now work as expected


mmackz:
Need to add in the OrderType, but I feel this should be quite trivial.

They also have stop-limit and stop-market orders. Should we just count limit and stop-limit as "limit" and likewise for stop-market and market.

mmackz:
Converted to use the $bitmask operator to unpack the packed values before comparisons. 

Everything should now work as intended 🎉

pnpm-lock.yaml
```diff
@@ -11213,7 +11259,7 @@ packages:
   /wide-align/1.1.5:
     resolution: {integrity: sha512-eDMORYaPNZ4sQIuuYPDHdQvf4gyCF9rEEV/yPxGfwPkRodwEgiMUUXTx/dex+Me0wxx53S+NgUHaP7y3MGlDmg==}
     dependencies:
-      string-width: 1.0.2
+      string-width: 4.2.3
```
mmackz, line 11262: 🤔

Quazia (approved):
Nice LGTM

PR #130, opened 2023-12-14
Fix(gmx): remove V1 support

Quazia:
I left the arrays and the constants in place just so it's easy to expand support in the future.


mmackz (approved):
Looks Good. 👍

PR #133, opened 2023-12-15
fix(zora): assure `universalMinter` is defined before using string methods

mmackz:
This PR fixes an issue where the plugin can will crash if `universalMinter` is undefined and the string method `.toLowerCase()` is called.


mmackz:
All the chains that we support are found in the `zoraUniversalMinterAddress` object, so this is a low priority fix. It will only crash if an unsupported chainId is used as the key in `zoraUniversalMinterAddress`

Quazia (approved):
LGTM

PR #136, opened 2023-12-16
feat(sushi): add support for sushi swap

mmackz:
## Integration of `sushi` Plugin for QuestDK

This PR introduces the integration of the `sushi` plugin for QuestDK.

### About The Project
Sushi is a popular ethereum based dex which is deployed on several L2 networks. It functions primarily as a decentralized exchange using an automated market maker (AMM) algorithm to facilitate efficient swapping of tokens.

### Implementation Details

This plugin supports several both V2 and V3 swaps on Sushi that are routed via the `RouteProcessor` or the legacy `V2Router` contracts. 

### Current Limitations

- `amountOut` may not be accurate due to the input data using `amountOutMin`, which takes into account the amount of slippage the user has set in the UI. 
- Swaps using the Trident Contract are not supported. (see readme for more info)


packages/sushi/src/contract-addresses.ts
```diff
@@ -0,0 +1,28 @@
...
+import { Chains } from './utils'
+
+export const CHAIN_TO_CONTRACTS: { [_chainId: number]: Address[] | undefined } =
+  {
+    [Chains.ETHEREUM]: [
+      '0x5550D13389bB70F45fCeF58f19f6b6e87F6e747d', // RouteProcessor3_2
+      '0xd9e1cE17f2641f24aE83637ab66a2cca9C378B9F', // V2Router02
+    ],
+    [Chains.OPTIMISM]: [
+      '0xEb94EcA012eC0bbB254722FdDa2CE7475875A52B', // RouteProcessor3_2
```
Quazia, line 11: No `V2Router02` on OP?

Quazia (approved):
LGTM - looks like the Trident contract is something they were trying to push and don't really support anymore so unlikely we'll need to support it down the line.

packages/sushi/src/contract-addresses.ts
mmackz: Not that I could find. It is not listed in the docs or on their SDK.

PR #137, opened 2023-12-17
feat(treasure): add support for token swaps on magicswap

mmackz:
## Integration of `treasure` swap plugin for QuestDK

This PR introduces the integration of the `treasure` plugin for QuestDK and implements the swap action type for tokens.

### About The Project
Treasure is an Arbitrum-based web3 gaming hub that has a large ecosystem of games, items and in-game ERC-20 tokens you can earn as you play.

### Implementation Details

This plugin adds support for token swaps done using the [magicswap](https://magicswap.lol/) interface. Magic swap uses the uniswap V2 router, so implementation was very straight forward. There is support for most of the tokens available of the platform excluding VEE token. All others can be selected as token in or out (MAGIC, ANIMA, ELM, GFLY)

### Current Limitations

- `amountOut` may not be accurate due to the input data using `amountOutMin`, which takes into account the amount of slippage the user has set in the UI.

### DIFF
--- a/.changeset/flat-adults-wash.md
+++ b/.changeset/flat-adults-wash.md
@@ -1,0 +1,6 @@
+---
+"@rabbitholegg/questdk-plugin-registry": minor
+"@rabbitholegg/questdk-plugin-treasure": minor
+---
+
+add support for treasure swap (magicswap)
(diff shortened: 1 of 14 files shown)


Quazia (approved):
LGTM - any reason we arn't supporting `Vee`? Just curious, don't think it's critical just wondering if you checked `coingecko` and found the other ones so I understand what the front-end support should look like here for these obscure tokens.

mmackz:
> LGTM - any reason we arn't supporting `Vee`? Just curious, don't think it's critical just wondering if you checked `coingecko` and found the other ones so I understand what the front-end support should look like here for these obscure tokens.

Checked all the token contracts against coingecko api, and VEE was the only one not found.

PR #144, opened 2023-12-24
feat(handlefi): add support for HandleFi swap plugin

mmackz:
## Integration of `handlefi` swap action plugin for QuestDK

This PR introduces the integration of the `handlefi` swap plugin for QuestDK.

### About The Project

HandleFi is a decentralized multi currency stablecoin protocol based on Arbitrum. They offer several products including perpetual swaps, convert, staking, and fxTokens.

### Implementation Details

There are several contracts that a swap may be routed through, making this plugin quite complex to implement. The majority of swaps will be routed via paraswap. fxTokens are routed through handlefis contracts or external curve/balancer pools.

### Current Limitations

- amountOut may not be accurate due to some methods using `amountOutMin` which takes into account the amount of slippage set.
- amountOut will not be considered if the swap is routed through the HPSM2 contract. 
- trades routed through curve V2 factory may have some false positives. Since there is not a really good way to tell what pool it is without having the contract address passed in as an argument. Because of this, false positives can occur if the tokenOut is selected as USDT and USDC.e is traded, and vice-versa.


packages/handlefi/src/input-filters.ts
```diff
@@ -0,0 +1,190 @@
...
+  PARASWAP_ABI,
+  V2_ROUTER_ABI,
+  HPSM2_ABI,
+  HLP_CURVE_V2_ABI,
+  HLP_BALANCER_ABI,
+  CURVE_FACTORY_ABI,
+} from './abi'
+import { WETH } from './constants'
+
+export function getParaSwapFilter(params: SwapActionParams) {
```
Quazia, line 14: This is super cool - this way we can just recycle the filter generators for a given protocol against aggregates! Good idea.

packages/handlefi/src/input-filters.ts
```diff
@@ -0,0 +1,190 @@
...
+  const { tokenIn, tokenOut, amountIn, amountOut } = params
+  const internalEthAddress = '0xEeeeeEeeeEeEeeEeEeEeeEEEeeeeEeeeeeeeEEeE'
```
Quazia, line 16: This has to be my least favorite way to represent ETH lol

packages/handlefi/src/input-filters.ts
```diff
@@ -0,0 +1,190 @@
...
+    $abi: PARASWAP_ABI,
+    $or: [
+      {
+        // simpleswap, directUniV3Swap, directCurveSwap
+        data: {
+          fromToken: tokenInUsed,
+          fromAmount: amountIn,
+          toAmount: amountOut,
+          toToken: tokenOutUsed,
+          partner: '0xFa2c1bE677BE4BEc8851D1577B343F7060B51E3A',
```
Quazia, line 30: What do you think about making this a param from the function call? I would either do that or hardcode it to a const.

packages/handlefi/src/input-filters.ts
```diff
@@ -0,0 +1,190 @@
...
+    ...tokenInput,
+  }
+}
+
+export function getCurveV2FactoryFilter(params: SwapActionParams) {
+  const { tokenIn, tokenOut, amountIn, amountOut } = params
+
+  // There isnt really a good way to tell which pool you are in without knowing the contract address of the pool
+  // i will always be fxUSD (0)
+  // j can either be FRAX, USDC.e, or USDT (may get false posiitves here)
```
Quazia, line 166: I think the false positives are fine - it's a pretty obscure edge-case. We'll find out if there's a way for the users to leverage this though, if there's a way they'll find it 😆

packages/handlefi/src/utils.ts
```diff
@@ -0,0 +1,79 @@
+import type { ActionParams, FilterOperator } from '@rabbitholegg/questdk'
+import { getAddress, type Address, type Hash } from 'viem'
+
+export enum Chains {
```
Quazia, line 4: Is this used anywhere?

Quazia (approved):
LGTM - Nice work on this one. This has gotta be one of the most complex plugins yet.

packages/handlefi/src/utils.ts
mmackz: Removed this since we already have a constant for arbitrum chain id.

packages/handlefi/src/input-filters.ts
mmackz: Very obscure. The fxUSD frax pool sees roughly 1 tx a day, and the 3pool sees maybe 1 per week.

Quazia:
Changes look good! Let's get this merged in.

PR #149, opened 2023-12-26
fix(uniswap): add sepolia op testnet to excluded networks

mmackz:
PR to exclude sepolia OP testnet from supported networks


jonathandiep (approved):
lgtm. We may also want to pin the Uniswap SDK package in the case that they add more networks

PR #150, opened 2023-12-27
fix(symbiosis): bump/pin sdk version and add support for new abi

mmackz:
This PR adds support for the contract changes made by symbiosis last week (new abi). I added new test using the newer testing pattern, so future updates should be more easy to implement.


packages/symbiosis/src/Symbiosis.test.ts
```diff
@@ -1,138 +1,79 @@
...
-        recipient,
+  describe('should return a valid list of tokens for each supported chain', async () => {
+    const chainIdArray = await getSupportedChainIds()
+    chainIdArray.forEach((chainId) => {
+      test(`for chainId: ${chainId}`, async () => {
+        const tokens = await getSupportedTokenAddresses(chainId)
+        const addressRegex = /^0x[a-fA-F0-9]{40}$/
+        expect(tokens).to.be.an('array')
+        expect(tokens).to.have.length.greaterThan(0)
+        expect(tokens).to.have.length.lessThan(100)
```
jonathandiep, line 69: Is 100 just an arbitrary limit that we set for now?

packages/symbiosis/src/Symbiosis.test.ts
```diff
@@ -1,138 +1,79 @@
...
-        destinationChainId,
-        tokenAddress,
-        amount: GreaterThanOrEqual(amount),
-        recipient,
+  describe('should return a valid list of tokens for each supported chain', async () => {
+    const chainIdArray = await getSupportedChainIds()
+    chainIdArray.forEach((chainId) => {
+      test(`for chainId: ${chainId}`, async () => {
+        const tokens = await getSupportedTokenAddresses(chainId)
+        const addressRegex = /^0x[a-fA-F0-9]{40}$/
```
jonathandiep, line 66: [nit] I would tend to use viem's `isAddress` to check if a string is a valid address
https://viem.sh/docs/utilities/isAddress.html#isaddress

BUT it does actually use this regex!
https://github.com/wevm/viem/blob/main/src/utils/address/isAddress.ts

packages/symbiosis/src/utils.ts
```diff
@@ -0,0 +1,63 @@
...
+  ARBITRUM_ONE = 42161,
+  AVALANCHE = 43114,
+  LINEA = 59144,
+  SCROLL = 534352,
+}
+
+interface Transaction {
+  chainId: number
+  from: Address
+  hash?: Hash
```
jonathandiep, line 23: Curious about why hash is optional here? I always thought transactions have a hash?

jonathandiep (approved):
lgtm, did leave a new nit comments!

packages/symbiosis/src/utils.ts
mmackz: I sometimes use simulated transactions if I cant find an actual transaction on the explorer.

The hash is not needed for the plugin, it is put there in case we want to find the transaction again in the explorer.

packages/symbiosis/src/Symbiosis.test.ts
mmackz: Learned something new! Awesome, I had no idea about this. Will change that out and start using it in all new plugins.

packages/symbiosis/src/Symbiosis.test.ts
mmackz: Ya, its arbitrary. Sometimes these sdks try to return massive lists of tokens. For example, uniswaps list has a few thousand, so we ended up just curating a list of popular tokens instead. So the test is basically just to alert us if an oversized list is being provided.

I didnt really put much though into the number used.

PR #152, opened 2023-12-27
fix(zora): fix overindexing issue with batchmint function

mmackz:
This pr fixes an overindexing issue when using the batchmint function.


Quazia (approved):
Nice - LGTM

PR #154, opened 2023-12-29
feat(camelot): add support for swaps on camelot V3

mmackz:
This PR introduces support for camelots V3 swap router into the `camelot` plugin.

V3 swap router contract - https://arbiscan.io/address/0x1f721e2e82f6676fce4ea07a5958cf098d339e18

This also fixes a bug with the current implementation where some token to ETH transactions were not being counted.


packages/camelot/src/Camelot.ts
```diff
@@ -6,105 +6,147 @@ import {
...
+      ],
+    },
     value: ethUsedIn ? amountIn : undefined,
     input: {
-      $abi: [...CAMELOT_ABI, ...PARASWAP_ABI],
       $or: [
         {
           // camelotV2 swap
-          to: recipient,
-          path: buildPathQuery(ethUsedIn ? Tokens.WETH : tokenIn, tokenOut),
```
mmackz, line 54: This was an issue with the current plugin implementation. if tokenOut is selected as ETH, it needs to be converted to WETH in the path array.

Quazia: Thanks for the callout

packages/camelot/src/Camelot.ts
```diff
@@ -6,105 +6,147 @@ import {
...
+          $abi: PARASWAP_ABI,
+          $or: [
+            {
+              // simpleswap, directUniV3Swap, directCurveSwap
+              data: {
+                fromToken: tokenInOrEth,
+                fromAmount: amountIn,
+                toAmount: amountOut,
+                toToken: tokenOutOrEth,
+                partner: '0x353D2d14Bb674892910685520Ac040f560CcBC06',
```
Quazia, line 105: Let's use a const for this

Quazia (approved):
LGTM

Quazia:
LGTM

PR #155, opened 2024-01-02
feat(mirror): add support for Mirror mint plugin

mmackz:
## Integration of `mirror` mint action plugin for QuestDK

This PR introduces the integration of the `mirror` mint plugin for QuestDK.

### About The Project

Mirror is a decentralized writing platform leveraging blockchain technology. It enables creators to monetize their work by turning their written content into collectible NFTs.

### Implementation Details

Each article on mirror is minted as an ERC-721 token and can be collected by making a simple mint transaction.

### Current Limitations

- Collecting Mirror Entries utilizes ERC-721 tokens, which increments the tokenId by 1 after every mint. Because of this, it is not practical to use the tokenId.
- There is a limit of 1 mint (collect entry) per account, so if the required amount is set higher than 1, the quest will be impossible to complete.


Quazia (approved):
LGTM - nice and easy one 👍

packages/mirror/src/test-transactions.ts
```diff
@@ -0,0 +1,55 @@
...
+    contractAddress: '0x189950164e777796CDF8844E030c300A01e65d1c',
+    recipient: '0xfFB3b6E273986C6adC31A1F0146b8ea69e3c306C',
+  },
+}
+
+export const passingTestCases = [
+  createTestCase(
+    OP_COLLECT_ENTRY,
+    'when collecting a mirror entry on optimism',
+  ),
```
mmackz, line 42: forgot to add the zora test

PR #156, opened 2024-01-02
Create vela guest

lovetrading10:


Quazia (changes requested):
Looks like a good start! Could you please add tests for the actions you added. I would look at our [open PR for the Vela swap](https://github.com/rabbitholegg/questdk-plugins/pull/129/files#diff-d4a7cab5303c3ad8954652ffe2ecd97ff9869823c213cb36cfa0d16dda11fde0) to see how the tests are setup.

Could you also revert the VS code settings changes please.

Once it's ready we're also [going to need a changeset](https://github.com/rabbitholegg/questdk-plugins#changesets--publishing) if you could add one.

I'll re-review once there are some tests to work with. Thanks!

packages/vela/src/Vela.ts
```diff
@@ -0,0 +1,83 @@
...
+  type SwapActionParams,
+  type StakeActionParams,
+  compressJson,
+} from '@rabbitholegg/questdk'
+import { type Address } from 'viem'
+import { VAULT_CONTRACT, CHAIN_TO_TOKENS, TOKENFARM_CONTRACT, VLP_CONTRACT, VELA_CONTRACT } from './contract-addresses'
+import { CHAIN_ID_ARRAY } from './chain-ids'
+import { TOKENFARM_ABI, TOKENFARM_ABI2, VAULT_ABI } from './abi'
+
+export const swap = async (
```
mmackz, line 13: You might just want to take this out and use our new derivatives action type instead of a swap. It would fit this action much better and have much better support on the front end once it (derivatives action) has been fully implemented.

packages/vela/src/Vela.ts
```diff
@@ -0,0 +1,83 @@
...
+): Promise<TransactionFilter> => {
+  const { chainId, contractAddress, tokenIn, amountIn, amountOut, deadline, recipient } = swap
+
+  return compressJson({
+    chainId,
+    to: contractAddress ?? VAULT_CONTRACT,
+    from: recipient,
+    input: {
+      $abi: VAULT_ABI,
+      a: amountIn,
```
mmackz, line 24: These values are packed, and will not match as you have them here.

I made a couple helper functions for unpacking the data that could help.

https://github.com/rabbitholegg/questdk-plugins/blob/f030d3e58474e46b0df73ba372009238797919b5/packages/vela/src/utils.ts#L68

packages/vela/src/Vela.ts
```diff
@@ -0,0 +1,83 @@
...
+      b: amountOut, 
+      c: deadline,
+    },
+  })
+}
+
+export const mint = async (
+  mint: SwapActionParams,
```
mmackz, line 32: Our mint action is mainly for NFTs and will not be able to work properly with amounts. 

I notice you are using SwapActionParams here, which wont play well with our frontend.

We could just use a swap action for this, but it might be a bit funky in regards to our frontend (The quest would be labeled as "Swap on Vela" and our users would get confused since your UI says Mint).

packages/vela/src/Vela.ts
lovetrading10: @mmackz 
I checked your derivatives action type, but I dont think it is not fit with our new position orderPacked.
it includes only two bigInt parameters, but we need three bigInt parameters.

packages/vela/src/Vela.ts
lovetrading10: Sorry but we cant use your way. 
image
You can check our newPositionOrderPacked function.

packages/vela/src/Vela.ts
lovetrading10: if possible, I hope you can create action type for vela.
Honestly, it doesn't look good to me.
But for following your sdk format, I am just using the actionType.

packages/vela/src/Vela.ts
mmackz: @lovetrading10 

I just finished submitting a pr to take care of the derivatives action.

https://github.com/rabbitholegg/questdk-plugins/pull/129


We were able to unpack the values using a bitmask.

packages/vela/src/Vela.ts
mmackz: Would just concentrate on the staking action for now, until we figure out a good way to match this action.

lovetrading10: sure.
thank you!

PR #158, opened 2024-01-04
feat(soundxyz): add support for Soundxyz mint plugin

mmackz:
## Integration of `soundxyz` mint action plugin for QuestDK

This PR introduces the integration of the `soundxyz` mint plugin for QuestDK.

### About The Project

Sound.xyz is the decentralized music marketplace where artists and their fans can connect in a more direct way. Every track on sound.xyz can be minted as an NFT with the proceeds going directly to the artist. 

### Implementation Details

Each sound on soundxyz is minted as an ERC-721 token and can be collected by making a simple mint transaction. 
There is two contracts that are used as entry points for a mint. Both use the same `mintTo` function. 
- [SuperMinter](https://optimistic.etherscan.io/address/0x0000000000cf4558c36229ac0026ee16d3ae35cd#code)
- [SuperMinterV2](https://optimistic.etherscan.io/address/0x000000000001a36777f9930aaeff623771b13e70#code)


packages/soundxyz/package.json
```diff
@@ -0,0 +1,49 @@
...
+    "rimraf": "^5.0.5",
+    "rome": "^12.1.3",
+    "ts-node": "^10.9.1",
+    "tsconfig": "workspace:*",
+    "typescript": "^5.2.2",
+    "vitest": "^0.33.0"
+  },
+  "dependencies": {
+    "@rabbitholegg/questdk": "2.0.0-alpha.28",
+    "viem": "^1.16.6"
```
Quazia, line 47: Let's pin this

pnpm-lock.yaml
```diff
@@ -11160,7 +11187,7 @@ packages:
   /wide-align/1.1.5:
     resolution: {integrity: sha512-eDMORYaPNZ4sQIuuYPDHdQvf4gyCF9rEEV/yPxGfwPkRodwEgiMUUXTx/dex+Me0wxx53S+NgUHaP7y3MGlDmg==}
     dependencies:
-      string-width: 4.2.3
+      string-width: 1.0.2
```
Quazia, line 11190: Whatsup with this?

Quazia (approved):
LGTM - I can circle back with the compression changes when I get the changes in upstream

pnpm-lock.yaml
mmackz: No idea tbh, I reset my dev environment and still get the same outcome.

packages/soundxyz/package.json
mmackz: pinned

PR #159, opened 2024-01-05
fix(traderjoe): fix issue with amount not working correctly using WETH

mmackz:
This pr fixes an issue with the current implementation where the amountIn is ignored when using WETH. 

Additional tests were also added to ensure the plugin will work as intended.


packages/traderjoe/src/abi.ts
```diff
@@ -0,0 +1,23 @@
+import { LBRouterV21ABI } from '@traderjoe-xyz/sdk-v2'
```
Quazia, line 1: Nice - glad you still used the import.

packages/traderjoe/src/test-transactions.ts
```diff
@@ -188,3 +188,20 @@ export const TOKENS_FOR_EXACT_TOKENS: TestParams<SwapActionParams> = {
     recipient: '0x22e798f9440f563b92aae24e94c75dfa499e3d3e',
   },
 }
+
+export const WETH_AMOUNT_IN: TestParams<SwapActionParams> = {
+  transaction: {
+    chainId: 42161,
+    from: '0xf44000901d480248b4dd1a388da2659cfdabec3f',
+    to: '0xb4315e873dbcf96ffd0acd8ea43f689d8c20fb30',
+    hash: '0xee5a4815403dfe13e487debbf65d74e16888a3d756f544a80edd8a9c4a4f7cc8',
```
Quazia, line 197: Nice - glad you grabbed one of the sybil hashes

Quazia (approved):
LGTM - thanks for getting this together so quick

PR #160, opened 2024-01-08
feat(vela): use $bitmask operator to unpack input values

mmackz:
merging latest changes into vela-swap branch


Quazia (approved):
This looks good but looking at it more I think we're going to need to refactor the 'handleBitmask' to support amount operators. Do you want to merge this upstream and then make the changes to the parent branch or wait until the SDK changes and just adjust in this branch? I think instead of an operator we just want a pre-processor and to not to the value comparison in the operator, just return the masked value. That way we can feed it into other functions and operators and it'll work with the amount operators. What do you think?

PR #161, opened 2024-01-09
feat(pendle): add support for Pendle swap plugin

tientran3:
### Integration of Pendle swap action plugin for QuestDK
---
This PR introduces the integration of the Pendle swap (swap YT only) plugin for QuestDK.

**Implementation Details**
Swap YT transactions will be sent to the Router contract. So we just need to compare the `to` address with Router contract and parse with the swap YT ABIs.
The filter will accept any token and amount with any market of supported chains.


packages/pendle/src/Pendle.ts
```diff
@@ -0,0 +1,37 @@
...
+} from '@rabbitholegg/questdk'
+import { type Address } from 'viem'
+import { SUPPORTED_CHAINS_ARRAY, type SupportedChainId } from './chain-ids'
+import { getRouterAddress } from './contract-addresses'
+import { SwapYTV3_ABI } from './abi'
+
+export const swap = async (
+  swap: SwapActionParams,
+): Promise<TransactionFilter> => {
+  const { chainId } = swap
```
Quazia, line 14: As it stands the only parameters you'll be able to create quests for would be chain ID, I'd recommend adding a token and token amount parameter as well.

packages/pendle/src/Pendle.test.ts
```diff
@@ -0,0 +1,56 @@
...
+  SWAP_EXACT_YT_FOR_TOKEN,
+} from './test-transactions'
+import { getRouterAddress } from './contract-addresses.js'
+import { SUPPORTED_CHAINS } from './chain-ids.js'
+import { SwapYTV3_ABI } from './abi.js'
+
+describe('Given the Pendle plugin', () => {
+  describe('When handling the swap', () => {
+    const validTransactions = [SWAP_EXACT_TOKEN_FOR_YT, SWAP_EXACT_YT_FOR_SY]
+    const invalidTransactions = [
```
Quazia, line 19: Generally an invalid transaction would be a transaction that has params that don't align with the filter being applied - what about these transactions makes them invalid for this context?

Quazia (changes requested):
Do you think it would be possible to add Amount and Token params into the filter? If you need help trying to figure out how to decode the tuples in the function calls we can definitely help. I'd also like to understand the failing/passing transactions as you have them right now. Can you explain better why some are considered valid/invalid?

packages/pendle/src/Pendle.test.ts
tientran3: Pendle recently released Router V3, and these transactions were sent to the old contract addresses. So I think they should be considered as invalid transactions

packages/pendle/src/Pendle.ts
leduythuccs: Hi, currently we are allowing all tokens, allow selecting specific token in or tokens out would require too much handling, for example:
- If the token in is normal ERC20 tokens, then the token out must be either YT or PT. 
- If the token in is YT, tokens out are either normal ERC20 tokens or PT
- if the token in is PT, tokens out are either normal ERC20 tokens or YT
- ... same for tokens out
Not to mention that we have multiple PT/YT tokens.

leduythuccs:
hi, after an internal discussion, we want to support the campaign that:
- Allow all token to **buy** our YT tokens.

So I don't think the token amount & select allowed tokens are needed. 

But if we need it, can we make an API call to our backend to get the list of YT tokens available?

Quazia:
@leduythuccs @tientran3 

Talked with the team - this should work for us to start and run some campaigns for you but we'll need the token and token amounts in order to add it to terminal.

What this means for now:
- We would be able to create campaigns and quests internally
- Third parties would not be able to permissionlessly deploy quests
- If you wanted to run quests you would need to ask us to start them

As long as this isn't an issue we're good to merge this in and start the integration testing process with our backend/indexer.

leduythuccs:
Hi, thank you for the reply,

That is not an issue at all!

About the list of the supported tokens, can we make an API call to our backend to fetch the list of supported tokens? Basically an API call to https://api-v2.pendle.finance/core/docs#/Assets/AssetsController_allAssets to fetch all the tokens that we supported.

If that is good we will update the PR accordingly

Quazia:
@leduythuccs 

Yeah absolutely no issue with making an API call - ideally if possible it'd be best to have it fallback to a safe default list if the API call fails but no reason not to have an API call here. It's actually probably better since we won't need to update it, a lot of projects just don't have good accessible endpoints for us to pull the token lists from which is why we usually fallback to hardcoded constants.

Do you want to add the API logic or just move forward as is?

leduythuccs:
Yes then let us go with the adding API call. We will implement a fallback too. Thank you a lot for your information!

tientran3:
I added `tokenIn`, `amountIn` and API call as well. Please help us to check. Thanks.

Quazia (approved):
LGTM - thanks for making those changes

PR #162, opened 2024-01-12
feat(mux): add support for Mux Protocol derivatives plugin

mmackz:
## Integration of `mux` derivative action plugin for QuestDK

This PR introduces the integration of the `mux` derivative plugin for QuestDK.

### About The Project

MUX Protocol is an arbitrum based decentralized perpetual aggregator. The aggregator is designed to offer traders the most optimal trading experience. 

### Implementation Details

Trades can be routed through 1 of 4 liquidity sources
- Mux Protocol
- GMX V1
- GMX V2
- Gains Network

### Current Limitations

- When routing through GMX, the recipient parameter is not available in the inputs. Trades routed this way will treat recipient as if it was undefined.
- Only DAI is allowed for trades routed through GAINS. There is no way to match the `token` parameter in the inputs. If the rest of the inputs match, it should be implied that they used DAI.
- There may be other ways a V1 GMX trade can be routed. From my own experience, it uses the same aggregator contract that gmx V2 trades are routed through, and uses the same function signatures. Is is possible something got missed though.


packages/mux/src/Mux.ts
```diff
@@ -0,0 +1,132 @@
...
+    : undefined
+
+  const getOrderType =
+    orderType === OrderType.Market
+      ? 0
+      : orderType === OrderType.Limit
+      ? { $gt: '0' }
+      : undefined
+
+  const gmxV2OrderFlag =
```
Quazia, line 47: I think the double ternary is a little hard to read - wdyt about a helper and maybe a switch statement or an if/else series of returns or something?

packages/mux/src/Mux.ts
```diff
@@ -0,0 +1,132 @@
...
+    chainId,
+  )
+
+  const tokenIn = token
+    ? token === zeroAddress
+      ? chainToWeth[chainId]
+      : token
+    : undefined
+
+  const getOrderType =
```
Quazia, line 40: Same comment re:double ternary

packages/mux/src/helpers.ts
```diff
@@ -0,0 +1,71 @@
...
+  chainId: number,
+  tokenAddress: Address,
+): Promise<string | undefined> => {
+  if (!tokenAddress) return undefined
+  const provider = CHAIN_ID_TO_PROVIDER[chainId]
+  if (!provider) {
+    throw new Error(`No provider found for chain ${chainId}`)
+  }
+
+  // cache assets for each chain to reduce api calls
```
Quazia, line 28: This won't persist - makes sense to cache the call but we'd need a way to persist the cache across runs. We only run plugins once per quest creation though so the frequency is so low it shouldn't matter.

packages/mux/src/helpers.ts
```diff
@@ -0,0 +1,71 @@
...
+
+  const tokenId = assets.find(
+    (asset) =>
+      asset.tokenAddress ===
+      (tokenAddress === zeroAddress
+        ? chainToWeth[chainId]
+        : getAddress(tokenAddress)),
+  )?.id
+
+  if (!tokenId) return 'ff' // this value doesnt exist, using instead of error
```
Quazia, line 47: If we return 'ff' the condition on :64 will presumably always fail - if we return '' it'll always pass what do you think is better in this context?

Quazia (approved):
Jeeeeez that's a beast of a filter. I'm still going to make those bitmask changes but it doesn't look like it matters for this one size there shouldn't be any `gte` `lte` etc in this one.

packages/mux/src/helpers.ts
mmackz: It should basically never reach here unless an invalid token is entered. It should always fail if it hits this condition imo.

Quazia: Sounds good to me

PR #163, opened 2024-01-12
feat(okutrade): add support for base network to okutrade swap plugin

mmackz:
This PR adds base network support for the OkuTrade swap plugin.


Quazia (approved):
LGTM

PR #164, opened 2024-01-12
Feat(gmx): add options

Quazia:


mmackz (commented):
Do we want this to be explicit?

packages/gmx/src/test-setup.ts
```diff
@@ -105,3 +107,53 @@ export const failingTestCasesV2 = [
   }),
   createTestCase(REDUCE_ORDER_V2, 'when wrong OrderType is used'),
 ]
+
+export const passingOptionsTestCases = [
+  createTestCase(MARKET_SHORT_ETH_ORDER_V2, 'when using a market order'),
+  createTestCase(LIMIT_LONG_ARB_ORDER_V2, 'when using a limit order'),
```
mmackz, line 113: would also suggest running some tests with some or all of the parameters undefined ie: any token quests

check line 74 for example

.gitignore
```diff
@@ -10,6 +10,7 @@ coverage
 dist
 node_modules
 tsconfig*.tsbuildinfo
+~
```
mmackz, line 13: curious what this is for

.vscode/settings.json
```diff
@@ -4,8 +4,8 @@
   "typescript.tsdk": "node_modules/typescript/lib",
   "typescript.enablePromptUseWorkspaceTsdk": true,
   "editor.codeActionsOnSave": {
-    "quickfix.rome": true,
```
mmackz, line 7: do we want this changed?

mmackz (approved):

packages/gmx/src/GMX.ts
```diff
@@ -90,6 +92,52 @@ export const swap = async (
...
+    },
+  })
+}
+
+export const getOrderType = (orderType: BoostOrderType | undefined) => {
+  switch (orderType) {
+    case BoostOrderType.Market:
+      return {
+        $or: [
+          { orderType: OrderType.MarketSwap },
```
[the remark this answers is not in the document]
Quazia: I think we're going to have less support pressure if we just accept it for the time being... we can remove it if it becomes an issue wdyt?

Quazia:
> Do we want this to be explicit?

What explicit?

.vscode/settings.json
Quazia: Yeah, I just didn't want that change tucked into an outside contributors changeset.

.gitignore
Quazia: Related to pnpm files getting loaded locally - didn't want to push them to remote.

mmackz:
> > Do we want this to be explicit?
> 
> What explicit?



> > Do we want this to be explicit?
> 
> What explicit?

the vscode setttings json

mmackz (approved):

mmackz:
LGTM

PR #166, opened 2024-01-14
feat(gmx): add support for limit swap

mmackz:
Our previous version of swap plugin only considered "market" swaps. This PR introduces support for both limit and market swap ordertypes.


packages/gmx/src/GMX.ts
```diff
@@ -65,7 +65,7 @@ export const swap = async (
                 numbers: {
                   minOutputAmount: amountOut,
                 },
-                orderType: OrderType.MarketSwap,
+                orderType: { $lte: OrderType.LimitSwap },
```
mmackz, line 68: tried `{ $or: [OrderType.LimitSwap, OrderType.MarketSwap] }` here, but it didnt seem to work properly. 

It was not filtering correctly, allowing any order type to pass the filter.

mmackz: 0 = MarketSwap
1 = LimitSwap

Quazia: This works fine but that's why[ I'm doing it like this](https://github.com/rabbitholegg/questdk-plugins/pull/164/files#diff-d80e7e326602ca01dfcd0f9ddaf9eb1acc908a0793337425d0ce0e0ea9d174c5R136).

PR #167, opened 2024-01-15
refactor(template): refactor project template

mmackz:
## Description

Refactor project template and provide a better description of how to adapt the filter function to use different action types that are available.


Quazia (approved):
LGTM

PR #172, opened 2024-01-19
feat(treasure): add mint plugin for TreasureTags

veganbeef:
This PR adds a mint transaction filter to the treasure plugin for TreasureTag mints.


packages/treasure/src/Treasure.test.ts
```diff
@@ -2,19 +2,66 @@ import { apply } from '@rabbitholegg/questdk/filter'
...
+      test('when minting a Treasure Tag through the proxy contract', async () => {
+        const filter = await mint(params)
+        expect(filter).to.deep.equal({
+          chainId: Chains.ARBITRUM_ONE,
+          to: params.contractAddress,
+          from: params.recipient,
+          input: {
+            $abi: MINT_TREASURE_TAG_ABI,
+            _registerArgs: {
+              owner: params.recipient,
```
veganbeef, line 29: is this `owner` param redundant given that it will always be the same as the `from` param?

packages/treasure/src/Treasure.ts
```diff
@@ -2,12 +2,31 @@ import {
...
+  const { chainId, contractAddress, recipient } = mint
+
+  return compressJson({
+    chainId,
+    to: contractAddress,
+    from: recipient,
+    input: {
+      $abi: MINT_TREASURE_TAG_ABI,
+      _registerArgs: {
+        owner: recipient,
```
veganbeef, line 24: is this `owner` param redundant given that it will always be the same as the `from` param?

Quazia: Yeah, either/or is fine

packages/treasure/src/Treasure.test.ts
```diff
@@ -2,19 +2,66 @@ import { apply } from '@rabbitholegg/questdk/filter'
...
+
   describe('When handling the swap action', () => {
     describe('should return a valid action filter', () => {
       test('when making a valid swap', async () => {
-        const { params } = passingTestCases[0]
+        const { params } = passingSwapTestCases[0]
         const filter = await swap(params)
         expect(filter).to.deep.equal({
-          chainId: 42161,
+          chainId: Chains.ARBITRUM_ONE,
```
Quazia, line 64: Nice cleanup ✨

Quazia (approved):
LGTM - if we expand out to have Bridgeworld or Smol probably makes sense to have them as seperate plug-ins.

PR #174, opened 2024-01-19
Updates/add mint vlp

lovetrading10:


packages/vela/src/contract-addresses.ts
```diff
@@ -4,11 +4,42 @@ import { Chains } from './utils'
...
+const ARBITRUM_DOGE = '0x5e90df307b07076c42b0516c683c45e10f362c18'
+const ARBITRUM_SHIB = '0x5033833c9fe8b9d3e09eed2f73d2aaf7e3872fd1'
 const ARBITRUM_ARB = '0x912ce59144191c1204e64559fe8253a0e49e6548'
+const ARBITRUM_SOL = '0xa26a142bb3e9d5503b882d17cff43a37b62ee1be'
+const ARBITRUM_MATIC = '0x561877b6b3dd7651313794e5f2894b2f18be0766'
+const ARBITRUM_AVAX = '0x71cb836342a9072d39fa100e2d44a982aff4b54d'
+const ARBITRUM_GBP = '0x1AE27D9068DaDf10f611367332D162d184ed3414'
+const ARBITRUM_EUR = '0x116172B2482c5dC3E6f445C16Ac13367aC3FCd35'
+const ARBITRUM_JPY = '0x95e0e6230e9E965A4f12eDe5BA8238Aa04a85Bc6'
+const ARBITRUM_AUD = '0x7E141940932E3D13bfa54B224cb4a16510519308'
```
mmackz, line 18: Most of these will not show up on our frontend, but I dont really see any issues with including all these tokens. 

@Quazia wdyt

packages/vela/src/test-transactions.ts
```diff
@@ -98,7 +100,59 @@ const ETH_STOPLIMIT: TestParams<OptionsActionParams> = {
   },
 }
 
-export const passingTestCases = [
+export const MINT_VLP: TestParams<OptionsActionParams> = {
```
mmackz, line 103: Even though this is not being used, it should not be using OptionsParams. Might be best to just remove this one for now.

mmackz:
Looking at the chat on telegram, it looks like you may be wanting to do Mint VLP.

Just to be clear, this pr for Staking VLP or VELA as shown on your frontend. 

Personally I think this is the better option for stake. Using the stake action for "mint/redeem" might be confusing for users, since the stake button is right next to it.

Screenshot from 2024-01-18 22-00-03

mmackz:
@lovetrading10 you will need to generate a changeset for this pr

run the command `pnpm changeset --filter=@rabbitholegg/questdk-plugin-vela`

lovetrading10:
> Looking at the chat on telegram, it looks like you may be wanting to do Mint VLP as the stake action. That would be fine, but we can only really support one stake action at a time, since they are so different.
> 
> Just to be clear, this pr for Staking VLP or VELA as shown on your frontend.
> 
> Personally I think this is the better option for stake. Using the stake action for "mint/redeem" might be confusing for users, since the stake button is right next to it.
> 
> Screenshot from 2024-01-18 22-00-03

@mmackz I am not going to do Minting VLP as Stake action. I separated Minting VLP and Staking VLP.
They are different actions.

mmackz (approved):

PR #175, opened 2024-01-19
feat(mirror): add support for legacy mints

mmackz:
Mirror blogs which were created past a certain point use an older function signature which was not recognized in the current implementation.

This PR adds an update to the ABI to support these mints, as well as some tests to ensure the filter is working as expected.


Quazia (approved):
LGTM - probably won't push out a release cycle just for this wdyt?

mmackz:
> LGTM - probably won't push out a release cycle just for this wdyt?

Sounds good to me.

PR #181, opened 2024-01-22
feature(treasure): add stake filter to treasure plugin

veganbeef:


packages/treasure/src/test-transactions.ts
```diff
@@ -24,6 +31,24 @@ export const MINT_TREASURE_TAG: TestParams<MintActionParams> = {
...
+    chainId: Chains.ARBITRUM_ONE,
+    contractAddress: MAGIC_STAKING,
+    tokenOne: MAGIC,
+    amountOne: '2000000000000000000',
+  },
+}
+
 export const EXACT_TOKENS_FOR_TOKENS: TestParams<SwapActionParams> = {
   transaction: {
     chainId: 42161,
```
jonathandiep, line 54: [nit] I think we might as well update the rest of the chainIds to also use `Chains.ARBITRUM_ONE`

veganbeef: Updated! I wasn't sure if there was a specific reason for the txns having hard-coded values

packages/treasure/src/test-transactions.ts
```diff
@@ -105,6 +148,7 @@ export const failingSwapTestCases = [
     tokenOut: GFLY,
   }),
   createTestCase(EXACT_TOKENS_FOR_TOKENS, 'when amountIn is insufficient', {
+    // should these amounts be hard-coded as values lower than the filters in the test params above?
```
Quazia, line 151: Yeah - I think one of the _only_ places hardcoding makes sense is for one-off failing test cases.

packages/treasure/src/Treasure.ts
mmackz, line 1: In the `getSupportedTokenAddresses` function, we are going to want to return a different list of tokens for swap and stake. Check out the paraswap package to see how this can be done.

mmackz (approved):

PR #183, opened 2024-01-23
feature(plugin): Adding support for Llama Vote plugin

gnomadic:


packages/llama/src/token-addresses.ts
```diff
@@ -0,0 +1,7 @@
+import { type Address } from 'viem'
+import { OPTIMISM_CHAIN_ID } from './chain-ids.js'
+
+export const LLAMA_TOKENS: Record<number, Address[]> = {
+  // TODO arguably we should support any token here?  it's a governance token right that's arbitraryq
```
Quazia, line 5: Yeah we need an `other` component and better interface validation. For now best to hardcode the addresses.

Quazia (changes requested):
Didn't look through the package file fully yet but here's the main feature review.

packages/llama/src/Llama.ts
```diff
@@ -0,0 +1,43 @@
...
+    // if the support param is a filter or undefined, then we can match either an approve or disapprove
+    abi = LLAMA_ABI_CAST_APPROVE.concat(LLAMA_ABI_CAST_DISAPPROVE)
+  }
+
+  return compressJson({
+    chainId: chainId, // The chainId of the source chain
+    to: project, // The contract address of the governance platform
+    input: {
+      $abi: abi,
+      ...(proposalId !== undefined && { actionInfo: { id: proposalId } }),
```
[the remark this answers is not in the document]
gnomadic: it wasn't working with `actionInfo.id: proposalId` because:


No value exists in scope for the shorthand property 'actionInfo'. Either declare one or provide an initializer.
You're trying to pass actionInfo as a key AND value to this object using a shorthand. You'll need to declare actionInfo as a variable first.

this overly complicated line is [from stack overflow](https://stackoverflow.com/questions/11704267/in-javascript-how-to-conditionally-add-a-member-to-an-object) but I tested it and it does work

packages/llama/src/LLama.test.ts
mmackz, line 1: Could you change filename to "Llama.test.ts"

packages/llama/src/chain-ids.ts
```diff
@@ -0,0 +1,3 @@
+export const OPTIMISM_CHAIN_ID = 10
```
mmackz: Would be preferable to use the Chains enum from the utils package

`import { Chains } from '@rabbitholegg/questdk-plugin-utils'`

packages/llama/src/chain-ids.ts
```diff
@@ -0,0 +1,3 @@
...
+
+export const CHAIN_ID_ARRAY = [OPTIMISM_CHAIN_ID]
```
mmackz: Llama will be on both Arbitrum and Optimism. We should add support for both.

packages/traderjoe/src/TraderJoe.test.ts
```diff
@@ -15,14 +10,18 @@ import {
...
+import { TOKENS_FOR_EXACT_TOKENS } from './test-transactions'
+import { Chains } from './utils'
 
 describe('Given the TraderJoe plugin', () => {
   describe('When handling the swap action', () => {
     describe('should return a valid action filter', () => {
       test('when swapping tokens', async () => {
         const { params } = TOKENS_FOR_EXACT_TOKENS
         const filter = await swap(params)
-        console.log(JSON.stringify(filter, null, 2))
```
mmackz, line 25: 👍 nice find

packages/llama/CHANGELOG.md
mmackz, line 1: We need to clear this changelog, since it contains the changes and versioning from the template file.

packages/llama/src/Llama.ts
```diff
@@ -0,0 +1,43 @@
...
+    },
+  })
+}
+
+export const getSupportedTokenAddresses = async (_chainId: number) => {
+  // Given a specific chain we would expect this function to return a list of supported token addresses
+  return LLAMA_TOKENS[_chainId]
```
mmackz, line 37: Im not sure if we need to return token addresses for a vote plugin 🤔. @Quazia wdyt? 

For tokens to work correctly here, the tokens need to be indexable on coingecko services, so we can obtain the icon and decimals.

https://github.com/rabbitholegg/rh/blob/main/src/services/internal/v1/tasks.ts#L11-L40

packages/llama/src/chain-ids.ts
gnomadic: I couldn't get this to work :/. I've been fighting build issues this whole PR and wasn't feeling like working through it 🙃

packages/llama/src/chain-ids.ts
gnomadic: I ended up adding every chain they support expect testnets!

packages/llama/src/Llama.ts
gnomadic: I ended up just setting all the factory contract addresses which won't work on coingecko eiterh 🙃  not sure if it's better to return an empty array?

packages/llama/src/Llama.ts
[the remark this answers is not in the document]
Quazia: Got it - this works for now.

I'll look at `compressJson` separately, the undefined collapses should percolate up.

packages/llama/src/Llama.ts
Quazia: Yeah probably best to leave it empty. The only param we could map to the supported token is the project address and that's probably best off as just an address entry vs a drop-down.

Quazia (approved):

packages/llama/CHANGELOG.md
gnomadic: should I just check it in empty with just the header?  is it populated by a github action?

mmackz: The header will be automatically generated.

packages/llama/src/chain-ids.ts
mmackz: Try just building the utils package.

`pnpm build --filter=@rabbitholegg/questdk-plugin-utils`

PR #192, opened 2024-01-25
feat(vela): add support for tp/sl orders.

mmackz:
This PR adds support for TP/SL orders. 

TP/SL orders use a function called newPositionOrderWithTPSL, which is not covered in the current implementation. This function does not use [packed parameters](https://arbiscan.io/tx/0x80228186ef0cc3ca15a532afb9dedb22be9aedf42eb3b2fef6cd0af0e6c507cc) like normal orders.

newPositionOrderWithTPSL: https://arbiscan.io/tx/0xa18ef3ec007d9427c93998676ee29757c58045dd6461e87dad0de49823e44357


packages/vela/src/Vela.ts
```diff
@@ -20,14 +31,28 @@ export const options = async (
       ? { $and: [getTokenPacked(token), getOrderTypePacked(orderType)] }
       : getTokenPacked(token) || getOrderTypePacked(orderType)
 
+  const amountTpslOrder = amount
+    ? [GreaterThanOrEqual(0), GreaterThanOrEqual(0), getAmount(amount)] // needs $nth operator
```
mmackz, line 35: Need to add an $nth operator to target nth item in array (this one being the 3rd item. There is a fourth item in this array if you are wondering why not use $last)

packages/vela/src/utils.ts
```diff
@@ -72,6 +72,35 @@ export function getOrderTypePacked(
   }
 }
 
+export function getOrderType(
+  orderType: OrderType | undefined,
+): { _orderType: undefined } | { $or: { _orderType: number }[] } | undefined {
```
Quazia, line 77: Could clean this up a bit with defined types - this reads a little gnarly but I'm not strongly convicted on that cleanup.

packages/vela/src/utils.ts
```diff
@@ -72,6 +72,35 @@ export function getOrderTypePacked(
...
+    },
+
+    [OrderType.Limit]: {
+      $or: [{ _orderType: 1 }, { _orderType: 3 }],
+    },
+  }
+  return orderTypeValues[orderType]
+}
+
+export function getAmount(amount: Amount): FilterOperator | undefined {
```
Quazia, line 91: Would be good to document why this is necessary in relation to their protocol and how they handle value with some function comments.

Quazia (approved):
LGTM - two small nits but feel free to ignore

PR #193, opened 2024-01-25
feat(utils): add default tokens for zkSyncEra

mmackz:
Pr adds a list of default tokens for zkSync Era chain.


packages/utils/src/constants/contract-addresses.ts
```diff
@@ -65,10 +65,17 @@ const arbitrumTokenAddresses: Address[] = [
   '0x82aF49447D8a07e3bd95BD0d56f35241523fBab1', // WETH
 ]
 
+const zkSyncEraTokenAddresses: Address[] = [
```
Quazia, line 68: We're going to have to get creative when we start naming these const values out. Might want to start using sub-files and importing them into this as we add more. Don't want this file to get too bloated.

PR #196, opened 2024-01-25
refactor(template): refactor template to introduce testing patterns and utilities

mmackz:


Quazia (approved):
LGTM - why swap vs bridge? Just cause it's more common? Mint might be another good one, those are really easy to do most of the time too.

mmackz:
> LGTM - why swap vs bridge? Just cause it's more common? Mint might be another good one, those are really easy to do most of the time too.

Mainly to support the documentation. Most of the examples I use are for swap actions. I choose swap because it is a common usecase and it is not as trivial as building a mint plugin.

PR #198, opened 2024-01-26
feat(boost): add support for boost mint

mmackz:
## Integration of `boost` Plugin for QuestDK

This PR introduces the integration of the `boost` plugin for QuestDK.

### About The Project
Boost Protocol is a token distribution network that revolutionizes marketing and user engagement in the web3 space. It rewards users for specific onchain actions, aligning incentives between projects, investors, and participants.

### Implementation Details

This plugin targets the minting of BoostPass NFTs. BoostPass NFTs are Soul-Bound ERC-721 tokens, which means that the tokenId increments after every mint. Because of this, we do not want to filter by tokenId. There is a limit of one mint per address.

To decode the `data_` input bytes, we used the $abiParams operator to decode the recipent parameter. The `data_` input contains the recipient of the boost pass, and the referrer. Only the recipient is relevant to the plugin.

### Contract Addresses
- https://sepolia.etherscan.io/token/0x9a618d6302f27cdbb97206ce269a31c1f7da3913 (boost pass)
- https://sepolia.etherscan.io/address/0x831c90d85c029208b18f74664ca6136573a47e9e (implementation)

### Current Limitations
- Any amounts will be ignored. It is not possible to mint more than one boost pass. 
- Any tokenIds will be ignored. TokenId is not applicable to ERC-721 tokens

### DIFF
--- a/packages/boost/src/Boost.ts
+++ b/packages/boost/src/Boost.ts
@@ -5,21 +5,27 @@ import {
 } from '@rabbitholegg/questdk'
 import { Chains } from '@rabbitholegg/questdk-plugin-utils'
 import { type Address } from 'viem'
+import {
+  BOOST_PASS_CONTRACT,
+  BOOST_PASS_ABI,
+  DATA_ABI_PARAMS,
+} from './constants'
 
 export const mint = async (
   mint: MintActionParams,
 ): Promise<TransactionFilter> => {
-  const {
-    chainId,
-    contractAddress,
-    recipient,
-  } = mint
+  const { chainId, contractAddress, recipient } = mint
 
   return compressJson({
     chainId,
-    from: recipient,
-    to: contractAddress,
-    input: {}, 
+    to: contractAddress ?? BOOST_PASS_CONTRACT,
+    input: {
+      $abi: BOOST_PASS_ABI,
+      data_: {
+        $abiParams: DATA_ABI_PARAMS,
+        recipient,
+      },
+    },
   })
 }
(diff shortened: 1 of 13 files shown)


packages/boost/src/Boost.ts
```diff
@@ -0,0 +1,40 @@
...
+): Promise<TransactionFilter> => {
+  const { chainId, contractAddress, recipient } = mint
+
+  return compressJson({
+    chainId,
+    to: contractAddress ?? BOOST_PASS_CONTRACT,
+    input: {
+      $abi: BOOST_PASS_ABI,
+      data_: {
+        $abiParams: DATA_ABI_PARAMS,
```
gnomadic, line 25: why is $abiParams instead of $abi?

packages/boost/src/test-transactions.ts
```diff
@@ -0,0 +1,36 @@
...
+  },
+  params: {
+    chainId: 11155111, // sepolia testnet
+    contractAddress: '0x9a618d6302f27cdbb97206ce269a31c1f7da3913',
+    recipient: '0x865c301c46d64de5c9b124ec1a97ef1efc1bcbd1',
+  },
+}
+
+export const passingTestCases = [
+  createTestCase(BOOST_PASS_MINT, 'when minting a boostpass'),
```
gnomadic, line 26: ohhh this is a nice helper!

gnomadic (approved):
🚢

PR #199, opened 2024-01-26
Fix(global): issue with questdk imports

Quazia:


packages/sushi/package.json
```diff
@@ -36,8 +36,6 @@
     "tsconfig": "workspace:*"
   },
   "dependencies": {
-    "@rabbitholegg/questdk": "2.0.0-alpha.31",
-    "@sushiswap/core-sdk": "1.0.0-canary.34",
```
gnomadic, line 40: a couple have the project as an import but that's not necessary right

gnomadic (approved):

packages/sushi/package.json
Quazia: wdym - this SDK is used for this package. Right now I have anything shared at root and anything only used by one project in that project. Could probably bubble them all up.

mmackz (approved):

PR #208, opened 2024-02-02
feat(creator):  MVP of plugin creator CLI

gnomadic:
this is a tool to scaffold out a new plugin


apps/create-plugin/package.json
Quazia, line 1: Can you rely on shared deps from root and pin all versions?

apps/create-plugin/src/index.ts
```diff
@@ -0,0 +1,141 @@
+import { Command } from 'commander'
+import packageJson from '../package.json'
+import { createPlugin, logBoostStars } from './builder'
+const _prompts = require('prompts')
+const figlet = require('figlet')
+// import { Chains } from '@rabbitholegg/questdk-plugin-utils'
```
Quazia, line 6: This should work now

apps/create-plugin/src/index.ts
```diff
@@ -0,0 +1,141 @@
...
+      if (invalidNameRegex.test(name)) {
+        return 'Please enter a name without special characters'
+      }
+      return true
+    },
+  },
+  {
+    type: 'multiselect',
+    name: 'chain',
+    message: 'What blockchains ars the project on? (multi-select)',
```
Quazia, line 89: ars -> are

package.json
```diff
@@ -1,6 +1,7 @@
 {
   "name": "questdk-plugins",
   "scripts": {
+    "create": "turbo run build  --filter=!@rabbitholegg/questdk-plugin-project && turbo run dev --filter=@rabbitholegg/create-plugin",
```
Quazia, line 4:
```suggestion
    "create": "pnpm build && pnpm run -r --filter=@rabbitholegg/create-plugin create",
```

apps/create-plugin/src/index.ts
```diff
@@ -0,0 +1,141 @@
...
+const _questions = [
+  {
+    type: 'text',
+    name: 'name',
+    message: 'What is the name of the project you are creating a plugin for?',
+    validate: (name: string) => {
+      if (name.length < 1) {
+        return 'Please enter a name'
+      }
+      const invalidNameRegex = new RegExp('[~"#%&*:<>?/\\{|}]+')
```
Quazia, line 79: I don't think this validation is tight enough - we could do something like "This is a BIG BAD PROJECT NAME aahahahahahaha" and that would be less than ideal. I think we want something more like this:
```suggestion
      const invalidNameRegex = new RegExp('"^(?:(?:@(?:[a-z0-9-*~][a-z0-9-*._~]*)?/[a-z0-9-._~])|[a-z0-9-~])[a-z0-9-._~]*$"')
```

That's just the package name validation regex, it might not be tight enough but at least it'll prevent breakage.

Quazia (changes requested):
Looks good, some small tweaks

apps/create-plugin/package.json
gnomadic: I think there were only 2 shared deps, and I thought they were pinned so I might not know what that means - is the ^<version> enough?

apps/create-plugin/src/index.ts
gnomadic: it does not :/

Screenshot 2024-02-07 at 6 52 44 PM

Quazia: This looks like a problem with the plugin creator packages build vs an issue with the import. I can fix it if you want but I want to rely on these shared utils where possible now that we have them.

gnomadic: I'll agree on reusing em - I tried - and I'll take you up on the fix, I'm not very experienced with typescript builds.

gnomadic: we could also resolve this one later it doesn't feel critical

apps/create-plugin/package.json
Quazia: [A caret will allow for floating versions](https://github.com/npm/node-semver?tab=readme-ov-file#caret-ranges-123-025-004) of at least patch and potentially minor. Ideally in this repo (and most of our repos) we pin our versions [pin here meaning that we depend on a _specific_ version vs a range of versions] and rely on dependabot to handle minor and patch versions. This ensure we get the benefit of a Github actions check flow to double check there are no breaking changes [even though there shouldn't be], and makes our overall package management a little easier to coordinate.

Quazia (approved):

PR #209, opened 2024-02-06
fix(tsconfig): update tsconfig.build files with explicit declarations path

ccashwell:
This fixes inconsistent turbo builds where the contents of `dist/types` folders is inconsistent, sometimes containing types and other times not. The package that threw the error was somewhat random due to turbo's indeterminate build order. Below is a sample error log:

```
src/Boost.ts(6,24): error TS7016: Could not find a declaration file for module '@rabbitholegg/questdk-plugin-utils'. '/Users/ccashwell/code/questdk-plugins/packages/utils/dist/cjs/index.js' implicitly has an 'any' type.
  Try `npm i --save-dev @types/rabbitholegg__questdk-plugin-utils` if it exists or add a new declaration (.d.ts) file containing `declare module '@rabbitholegg/questdk-plugin-utils';`
src/test-transactions.ts(6,8): error TS7016: Could not find a declaration file for module '@rabbitholegg/questdk-plugin-utils'. '/Users/ccashwell/code/questdk-plugins/packages/utils/dist/cjs/index.js' implicitly has an 'any' type.
  Try `npm i --save-dev @types/rabbitholegg__questdk-plugin-utils` if it exists or add a new declaration (.d.ts) file containing `declare module '@rabbitholegg/questdk-plugin-utils';`
```

NB: It is possible there's an alternative resolution path that doesn't require repetition, but I wasn't able to get consistent behavior with only changes to the base config.

### DIFF
--- a/packages/across/tsconfig.build.json
+++ b/packages/across/tsconfig.build.json
@@ -9,6 +9,8 @@
     "scripts/**/*"
   ],
   "compilerOptions": {
+    "declaration": true,
+    "declarationDir": "./dist/types",
     "resolveJsonModule": true,
     "sourceMap": true,
     "rootDir": "./src"
--- a/packages/arbitrum/tsconfig.build.json
+++ b/packages/arbitrum/tsconfig.build.json
@@ -9,6 +9,8 @@
     "scripts/**/*"
   ],
   "compilerOptions": {
+    "declaration": true,
+    "declarationDir": "./dist/types",
     "resolveJsonModule": true,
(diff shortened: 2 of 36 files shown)


Quazia (approved):
Thank you so much for this solve 🙏  LGTM

There's no changeset but I reckon we don't need to cut a release to tweak the build - can just rely on subsequent releases to include these changes wdyt

ccashwell:
agreed, no release required here, next actual change is fine

PR #214, opened 2024-02-09
fix(gmx): remove swaps from getOrderType function

mmackz:
Fixes and issue where basic swaps are passing the "open a position" boosts.


Quazia (approved):
LGTM :shipit:

PR #217, opened 2024-02-15
Feat: add mint transaction intents to plugins

Quazia:
This adds `getMintIntent` to Zora, Mirror, and sound. It also adds `getTxIntent` to the registry. Using these we can translate `ActionIntentParams` into `TransactionRequest` objects. These objects can then be signed without the front-end needing to know the specific ABI related to a protocol.


packages/utils/src/types/actions.ts
```diff
@@ -0,0 +1,150 @@
+import { type TransactionRequest } from 'viem'
+import type {
+  FilterOperator,
+  TransactionFilter,
+  PluginActionNotImplementedError,
+} from '@rabbitholegg/questdk'
+import { type Address } from 'viem'
+import type { MintIntentParams } from './intents'
+
+export type SwapActionParams = {
```
gnomadic, line 10: might be missing something but why is this duplicated into plugins?  Looking at sibling PR for the questdk, and these types were not removed?

gnomadic (approved):

PR #221, opened 2024-02-20
feat(builder): dynamic questions based on actionType

mmackz:
This updates the plugin builder to ask dynamic questions based on the action type and builds the test-transactions.ts file based on input given by the user.

It uses viem to grab the details of the transaction and the tokendetails (symbol/decimals) and uses this info to build test cases.

After this update, the user should only need to create the transaction filter and generate a changeset.

includes #220


Quazia:
@mmackz Is this good for review? If so I'll try and get it tackled tomorrow depending how I'm doing with the dynamic names work I'm on right now.

mmackz:
> @mmackz Is this good for review? If so I'll try and get it tackled tomorrow depending how I'm doing with the dynamic names work I'm on right now.

Yes please 🙏

mmackz:
Builder Bugs 

- Chains should only be imported if at least one chain is selected (Project.ts)
-   GreaterThanOrEqual should only be imported if there is a valid amount. (test-transactions.ts)
- Some issues with GreaterThan or Equal on vote action and others with amounts that are not swap or bridge.

gnomadic (approved):

apps/create-plugin/template/src/test-transactions.ts.t
```diff
@@ -0,0 +1,65 @@
...
+  Chains{{/eq}}
+} from '@rabbitholegg/questdk-plugin-utils'
+{{else}}
+import { type TestParams } from '@rabbitholegg/questdk-plugin-utils'
+{{/if}}
+
+{{#unless tx.length}}
+export const {{uppercase actionType}}_TEST: TestParams<{{capitalize actionType}}ActionParams> = {
+  transaction: {
+    chainId: '',
```
gnomadic, line 23: can we set these with valid typed default values so it compiles if the user doesn't provide a tx?

packages/apples/README.md
```diff
@@ -0,0 +1,15 @@
+## Apples Plugin for Boost
+
+### New Plugin TODO list
```
gnomadic, line 3: whoops also delete this test output :)

apps/create-plugin/template/src/test-transactions.ts.t
mmackz: FIxed!

PR #228, opened 2024-02-22
fix(zora): add batchmint support

mmackz:
fixes issue with batchmint not being recognized on arbitrum


Quazia (approved):
Weird the tslib version rolled back but looks consistent.

LGTM

PR #236, opened 2024-02-23
Feat(sdk): dynamic names

Quazia:
Right now I have `getDynamicNameParams` taking in:
```
getDynamicNameParams?: (
    params: DisctriminatedActionParams,
    metadata: Record<string, unknown>,
  ) => Promise< Record<string, unknown>> ;
  ```
  A lot of the helpers right now are passing in the action type and the data blob seperately, preventing discrimination. For this helper I combined them and I want to go back, overload, then deprecate the helpers [that don't do this currently](https://github.com/rabbitholegg/questdk-plugins/blob/d3114da6a9614e397cacd0228e039cd63c13ebf3/packages/registry/src/index.ts#L157-L161)
  I considered doing the metadata with a stricter type, but basically every field is optional. We could do another discriminated parameter and discriminate on the project. Happy to implement that.
  
  I also played around with retrieving the metadata/params in this function but I think it's better if there's no requests fired from this helper. I think at the `questdk` level I'll leverage these helpers and expose a helper that just takes in a UUID and passes back the name with all of the requests out to the API at the `questdk` level.
  
  
  Another big change to signal towards is I added {originAuthor} and {originTargetImage} to the values object; some of these will be empty for some protocols.
  
  ```
  // Example Message: 
  // const messages = {
  //   en: {
  //     'boost.mint':
  //       '{actionType} {originQuantity} {originTargetImage} {originTarget} {originCollection} {originAuthor} {projectImage} {project} {originNetwork}',
  //     'mint.from': 'from the {collection} collection',
  //     'mint.anything': 'anything',
  //     exactly: 'exactly {amount}',
  //     'on.network': 'using {network}',
  //   },
  // };
  
  ```

### CODE REVIEW

File: apps/questdk/src/quests/fetchQuestData.live.test.ts
Comment (jonathandiep) [inline apps/questdk/src/quests/fetchQuestData.live.test.ts:1]: We'll do it live!!

File: apps/questdk/src/quests/fetchQuestData.test.ts [hunk 1]
```diff
@@ -0,0 +1,68 @@
...
+      questAddress: '0xabcdef1234567890',
+      totalParticipants: 100,
+      numberMinted: 50,
+      receiptsMinted: 50,
+      participants: [{ account: '0x1234567890abcdef', amount: '100' }],
+      questFactoryAddress: '0xabcdef1234567890',
+      queued: false,
+      isPublic: null,
+      actionParams: {} as any, // Fill with correct test data
+    }
```
jonathandiep, line 50: [nit] Not needed now, but I think would be great to have this in a mock helper folder so we can reuse this where needed

apps/questdk/src/tokens/nft.ts
```diff
@@ -0,0 +1,60 @@
...
+    })
+
+    // If the call succeeds, it's an ERC721 contract
+    return fetchERC721Metadata(client, contractAddress, tokenId)
+  } catch (error) {
+    // If the call fails, try to call the ERC1155 balanceOf function
+    await client.readContract({
+      address: contractAddress as Address,
+      abi: [
+        'function balanceOf(address account, uint256 id) external view returns (uint256)',
```
gnomadic, line 29: what's funny is that this could still fail 🙃 I've seen reject when tokenId doesn't exist

gnomadic:
does the client calling in always know the uuid of the quest?  I assume so

gnomadic (approved):

apps/questdk/src/tokens/nft.ts
Quazia: Yeah, it's not perfect. If the `tokenId` doesn't exist though we can't return metadata for it so 🤷

Quazia:
@gnomadic For now our clients do, and most of our partners do (this is the primary identifier) ideally in the future we can switch from UUID -> address as the PK for a boost. I want to expose UUID entry points at the questdk level for most of these getters and in the future we can overload it with whatever else but I want to avoid having third parties need to parse our or understand our data model too much to get _exactly what they need_ from the helpers where/when they need it.

PR #237, opened 2024-02-26
Calculate `projectFees` for mint action type

jonathandiep:
This PR adds the `projectFees` for Zora and Mirror mints

Will follow up with Sound

### DIFF
--- a/packages/mirror/src/Mirror.ts
+++ b/packages/mirror/src/Mirror.ts
@@ -1,17 +1,25 @@
 import {
-  type TransactionFilter,
   type MintActionParams,
+  type TransactionFilter,
   compressJson,
-} from '@rabbitholegg/questdk'
-import { type Address, encodeFunctionData, type TransactionRequest } from 'viem'
-import { COLLECT_ENTRY_ABI } from './abi'
-import { Chains } from './utils'
-import type { MintIntentParams } from '@rabbitholegg/questdk-plugin-utils'
+} from "@rabbitholegg/questdk";
+import {
+  type MintIntentParams,
+  chainIdToViemChain,
+} from "@rabbitholegg/questdk-plugin-utils";
+import {
+  type Address,
+  type TransactionRequest,
+  encodeFunctionData,
+} from "viem";
+import { http, createPublicClient } from "viem";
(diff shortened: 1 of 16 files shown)


packages/utils/src/index.ts
```diff
@@ -28,11 +33,11 @@ export type {
...
   FilterOperator,
   TransactionFilter,
-} from './types'
+} from "./types";
 
 export {
   ActionType,
   OrderType,
-} from './types'
+} from "./types";
```
jonathandiep: Used the VSCode Rome extension to format which explain lots of the formatting changes here. Ended up disabling since I didn't want formatting to make this PR look confusing on other existing files

We should probably discuss formatting later too

https://marketplace.visualstudio.com/items?itemName=rome.rome

packages/utils/src/index.ts
```diff
@@ -1,5 +1,10 @@
-export { CHAIN_TO_TOKENS, Chains } from './constants'
-export { createTestCase, type TestCase, type TestParams } from './helpers'
+export { CHAIN_TO_TOKENS, Chains } from "./constants";
+export {
+  createTestCase,
+  type TestCase,
+  type TestParams,
+  chainIdToViemChain,
```
jonathandiep, line 6: Only actual diff is the `chainidToViemChain` helper

packages/mirror/src/Mirror.ts
```diff
@@ -44,6 +48,102 @@ export const getMintIntent = async (
...
+        type: 'function',
+      },
+    ],
+    functionName: 'treasuryConfiguration',
+  })
+
+  // get feeConfiguration address
+  const feeConfiguration = await client.readContract({
+    address: treasuryConfiguration,
+    abi: [
```
Quazia, line 84: Same here - you can even drop both of them in the same ABI like `MIRROR_FEE_ABI` or something.

jonathandiep:
Latest commit shows that I added the changeset, but doesn't seem to pop up in this PR 🤔

packages/utils/src/index.ts
Quazia: I have a linear ticket captured to migrate to Biome, rome is deprecated.

PR #242, opened 2024-02-27
fix(woofi): update contract addresses

mmackz:
- refactor to use utils package
- update contract addresses to restore plugin functionality
- add test to ensure plugin works with new changes


packages/woofi/src/contract-addresses.ts
```diff
@@ -1,14 +1,14 @@
...
-  [Chains.OPTIMISM]: '0xEAf1Ac8E89EA0aE13E0f03634A4FF23502527024',
-  [Chains.POLYGON_POS]: '0x817Eb46D60762442Da3D931Ff51a30334CA39B74',
-  [Chains.ZKSYNC_ERA]: '0xfd505702b37Ae9b626952Eb2DD736d9045876417',
-  [Chains.BASE]: '0x27425e9FB6A9A625E8484CFD9620851D1Fa322E5',
-  [Chains.ARBITRUM_ONE]: '0x9aEd3A8896A85FE9a8CAc52C9B402D092B629a30',
-  [Chains.AVALANCHE]: '0xC22FBb3133dF781E6C25ea6acebe2D2Bb8CeA2f9',
+  [Chains.ETHEREUM]: '0x4c4AF8DBc524681930a27b2F1Af5bcC8062E6fB7',
+  [Chains.OPTIMISM]: '0x4c4AF8DBc524681930a27b2F1Af5bcC8062E6fB7',
+  [Chains.POLYGON_POS]: '0x4c4AF8DBc524681930a27b2F1Af5bcC8062E6fB7',
+  [Chains.ZK_SYNC_ERA]: '0x09873bfECA34F1Acd0a7e55cDA591f05d8a75369',
```
Quazia, line 8: Wonder why it's different only on ZKSYNC

Quazia (approved):
LGTM - need a changeset still you're pulling a me 😅

packages/woofi/src/contract-addresses.ts
mmackz: [There is a few networks ](https://learn.woo.org/v/woofi-dev-docs/references/readme)

It is different on the 2 zk networks, zksync and polygonzk. Not sure why that would be.

Quazia: Interesting looks like address derivation is slightly different there:

https://docs.zksync.io/build/developer-reference/differences-with-ethereum.html#evm-instructions

PR #247, opened 2024-02-27
Fix: getMintIntent export

Quazia:


mmackz:
LGTM

Does this need a changeset?

Quazia:
@mmackz yes gonna add one but I think I might need to massage some of the types first

PR #253, opened 2024-02-29
feat: add jojo stake plugin

almoapple:
jojo docs: https://about.jojo.exchange/jojo-1/jojo-exchange/overview
jojo app: https://app.jojo.exchange


packages/jojo/src/contract-addresses.ts
```diff
@@ -0,0 +1,8 @@
+import type { Address } from 'viem'
+
+export const FUNDING_RATE_ARBITRAGE =
+  '0x0174f6d813df42c986adf75ec473a0162faafcda'
+export const TOKEN_USDC = '0xaf88d065e77c8cC2239327C5EDb3A432268e5831'
+export const TOKEN_JUSD = '0xc3d8d1A7aC1C1924dCc742EC6cA62A775814a1bb'
+
+export const STATE_TOKENS: Address[] = [TOKEN_USDC, TOKEN_JUSD]
```
mmackz, line 8: You only should return USDC since it is the only token used as an input for staking. 

Is it STAKE_TOKENS or STATE_TOKENS

almoapple: Thanks for your advice, have done.

mmackz (approved):

PR #254, opened 2024-02-29
Chore: bump pnpm 8.15.3

Quazia:


jonathandiep (approved):
Are we going to need to change inside the `package.json` for each plugin?

Quazia:
There shouldn't be any inside of the individual packages, but now that I'm looking there are. So I'll delete them at the sub package level and then we should be good. Good catch @jonathandiep  thank you

PR #261, opened 2024-03-01
Fix(zora): update mint to new mint ABI

Quazia:
The ABI on new mints is slightly different. When simulating it's important that the address creating the transaction has enough funds and is eligible to mint.


packages/zora/src/Zora.ts
```diff
@@ -97,8 +100,15 @@ export const getMintIntent = async (
 ): Promise<TransactionRequest> => {
   const { contractAddress, tokenId, amount, recipient } = mint
   let data
-  if (tokenId !== 0) {
-    const mintArgs = [recipient, tokenId, amount, zeroHash]
+  if (tokenId !== null && tokenId !== undefined) {
+    const msgArgs = '0x' + recipient.slice(2).padStart(64, '0')
```
jonathandiep, line 104: [nit] I think we can use viem's `pad` with size = 64 to do this

https://viem.sh/docs/utilities/pad#size

packages/zora/src/Zora.test.ts
```diff
@@ -242,3 +232,23 @@ describe('Given the getProjectFee function', () => {
...
+      contractAddress: '0xc53c050131a3507d51d014445f666f4c3a1a2c24',
+      tokenId: 1, // not 0
+      amount: BigInt(1),
+      recipient: '0x742d35Cc6634C0532925a3b844Bc454e4438f44e',
+    }
+    const value = parseEther('0.000777')
+    const account = '0xE4eDb277e41dc89aB076a1F049f4a3EfA700bCE8'
+
+    const result = await simulateMint(mint, value, account)
+    console.log(result)
```
jonathandiep, line 249: Can remove this stray log

jonathandiep (approved):
lgtm, but left a couple of nits!

mmackz (approved):

mmackz:
Looks good. Needs  a changeset

packages/zora/src/Zora.ts
Quazia: Cool - will do

PR #270, opened 2024-03-02
feat(builder): use github actions to confirm and upload plugin details to db

mmackz:
Details
------
------
Issue
-----

------
Testing
------

I have done a lot of testing locally.


mmackz:
When no packages are found it should successfully pass the action.

CleanShot 2024-03-02 at 13 12 20@2x

PR #271, opened 2024-03-03
feat: use github actions to confirm and upload plugin details to db

mmackz:
Details
------

[Backend Changes](https://github.com/rabbitholegg/rh/pull/2405)

------
Issues
-----

- ~~When updating the config file with a new image url, it wont change the image on the S3 bucket unless you use a different name.~~ Image updates do work, the old image is just cached for a while. 😀

------
Testing
------

I have done **a lot** of testing locally.

**Passes**
  - [x] when all details are valid in new plugin config file
  - [x] when all details are valid on updated config file
  - [x] when all details are valid, adding a new config file to existing package
  - [x] when making non-config related changes (passes with "no changes found")
  - [x] when multiple valid tasks are provided

**Fails**
  - [x] when mandatory details are missing
  - [x] when image is too big
  - [x] when image url is not an image
  - [x] when icon or applink is not a url
  - [x] when a new package is detected and no config file is found
  - [x] when adding an invalid config file to an existing package
  - [x] when updated an existing config file using invalid details


.github/scripts/registerNewPlugins.ts
```diff
@@ -0,0 +1,84 @@
...
+const _utils = require("./utils");
+
+async function sendPluginDetailsToAPI(detailsPath: string): Promise<void> {
+  const fileContents = await _fs.readFile(detailsPath, "utf8");
+  const details = _yaml.load(fileContents);
+  const { project, task } = details;
+
+  // send details to staging API
+  const { data: stagingData } = await _axios.post(
+    "https://api-staging.boost.xyz/plugins/add-project",
```
mmackz, line 13: save these in env variable?

mmackz:
Pretty confident in how it operates. I tested this for many hours.

Here is the logic.

- Compares to main to see if there are any new package directories. If found, a valid `plugin-details.yml` must be provided.
- If a new `plugin-details.yml` file is found in an existing directory, it must be validated
- If any existing `plugin-details.yml` file is edited, it must be validated. Details will be compared to the current entry in the db, and any changes will be saved. This allows projects to update their icons and applink by submitting a pr.

mmackz:
One thing I have not considered is how to handle plugins with multiple tasks (actions)

mmackz:
I think we need to come up with better logic for what a plugin is, and what folders need to have a plugin config file. Currently this just looks for new folders in the packages directory and assumes they are new plugins, forcing them to have a plugin details config.

However, not all folders in this directory are going to be plugins (see ts-config, utils, registry) 🤔

mmackz:
I tried to merge this back into my test branch and it somehow ended up bringing over the comment history 😢

PR #272, opened 2024-03-04
feat(sound): add support for targeted mints of limited sounds

mmackz:
- maps the `tokenId` to the `tier` property which allows boost creators to target mints of limited edition sounds

- since all sound mints are erc-721, the tokenId is not otherwise used.

- May be some UX issues if we do not communicate that limited edition mints will pass.


Quazia (approved):
LGTM - they'll be pumped about this we should let coop records know we can support this now too.

PR #275, opened 2024-03-06
fixed token metadata functions

veganbeef:
This PR fixes the token metadata functions (see [this slack thread](https://rabbithole-tam3602.slack.com/archives/C0320B5C8UB/p1709588524518149) for context)

### DIFF
--- a/apps/questdk/src/tokens/erc1155.ts
+++ b/apps/questdk/src/tokens/erc1155.ts
@@ -15,15 +15,37 @@ export async function fetchERC1155Metadata(
   contractAddress: string,
   tokenId: number,
 ): Promise<any> {
-  const tokenURI: string = await (client.readContract({
+  let tokenURI: string = await (client.readContract({
     address: contractAddress as Address,
     abi: [
-      'function uri(uint256 tokenId) external view returns (string memory)',
+      {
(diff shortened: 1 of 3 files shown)


apps/questdk/src/tokens/erc1155.ts
```diff
@@ -15,15 +15,37 @@ export async function fetchERC1155Metadata(
...
+        type: 'function',
+      },
     ],
     functionName: 'uri',
-    args: [tokenId],
+    args: [BigInt(tokenId)],
   }) as Promise<string>)
 
+  if (tokenURI.startsWith('ipfs://')) {
+    const ipfsIdentifier = tokenURI.split('/')[2]
```
jonathandiep, line 46: Similarly to the other repo, we may want to grab the entire path (index 2 and beyond!)

Quazia (approved):
Thanks! This looks good. I'll add some live tests and move the ABIs into the utils, hopefully this covers all the launch cases at least though.

apps/questdk/src/tokens/erc1155.ts
veganbeef: good call! fixed

PR #280, opened 2024-03-07
refactor(symbiosis): remove symbiosis-sdk

mmackz:
Removes the dependency for symbiosis-sdk and replaces it with a mapping of chains to router contracts.


pnpm-lock.yaml
```diff
@@ -219,7 +219,7 @@ importers:
         version: 2.0.4(sinon@17.0.1)
       '@connext/smart-contracts':
         specifier: ^2.0.0
-        version: 2.0.0(ethers@5.2.0)
+        version: 2.0.0(ethers@5.7.2)
```
Quazia, line 222: Too bad ethers is still being used here

Quazia (approved):
LGTM thank you - should be able to re-enable the plugin after this

PR #284, opened 2024-03-08
Fix(sound): adjust fee calc to use schedules differently

Quazia:
We now infer the schedule from the next schedule and infer the tier from the token id.


packages/soundxyz/src/Soundxyz.ts
```diff
@@ -137,27 +137,26 @@ export const simulateMint = async (
...
     transport: http(),
   })
 
-  const mintInfoList = (await client.readContract({
+  const nextSchedule = (await client.readContract({
     address: SUPERMINTER_V2,
     abi: MINT_INFO_LIST_ABI,
-    functionName: 'mintInfoList',
-    args: [contractAddress],
-  })) as MintInfoList
```
jonathandiep, line 152: Since we no longer use `MintInfoList` can you remove the `MintInfo` and `MintInfo` types?

Or can just delete `questdk-plugins/packages/soundxyz/src/types.ts`

packages/soundxyz/src/constants.ts
```diff
@@ -65,6 +65,30 @@ export const SUPERMINTER_ABI = [
 ]
 
 export const MINT_INFO_LIST_ABI = [
```
jonathandiep, line 67: Lets rename this variable and delete `mintInfoList` abi

packages/soundxyz/src/Soundxyz.ts
```diff
@@ -137,27 +137,26 @@ export const simulateMint = async (
...
-
-  const mintInfo = mintInfoList[0]
+    functionName: 'nextScheduleNum',
+    args: [contractAddress, tier],
+  })) as bigint
```
jonathandiep, line 153: For some reason I think this is being cast as `number` instead of `bigint`

Lets try to use `number` and then cast into a `BigInt` below in the args

packages/soundxyz/src/Soundxyz.ts
jonathandiep, line 1: When testing the `mintIntent`, I think we are also going to need to make an update on the `scheduledNum`
CleanShot 2024-03-08 at 16 51 13@2x

packages/soundxyz/src/Soundxyz.test.ts
```diff
@@ -159,5 +161,8 @@ describe('getProjectFees', () => {
     const fee = await mockFns.getProjectFees(mintParams)
     expect(getProjectFeesSpy).toHaveBeenCalledWith(mintParams)
     expect(fee).toEqual(BigInt('777000000000000'))
+
+    // const projectFees = await getProjectFees(mintParams)
```
Quazia, line 165: Declutter

packages/soundxyz/src/Soundxyz.ts
```diff
@@ -96,20 +127,29 @@ export const simulateMint = async (
...
+  const { contractAddress, recipient, tokenId, amount } = mint
   const _client =
     client ||
     createPublicClient({
       chain: chainIdToViemChain(mint.chainId),
       transport: http(),
     })
 
+  const tier = tokenId ?? 0
+  const quantity = amount ?? 1
```
Quazia, line 139: Nice catch

Quazia (commented):
LGTM but I can't approve cause I'm author /ded

jonathandiep (approved):

PR #285, opened 2024-03-09
feat: use github actions to confirm and upload plugin details to db 

mmackz:
Details
------

[Backend Changes](https://github.com/rabbitholegg/rh/pull/2405)

------
Issues
-----

- ~~When updating the config file with a new image url, it wont change the image on the S3 bucket unless you use a different name.~~ Image updates do work, the old image is just cached for a while. 😀

------
Testing
------

I have done **a lot** of testing locally.

**Passes**
  - [x] when all details are valid in new plugin config file
  - [x] when all details are valid on updated config file
  - [x] when all details are valid, adding a new config file to existing package
  - [x] when making non-config related changes (passes with "no changes found")
  - [x] when multiple valid tasks are provided

**Fails**
  - [x] when mandatory details are missing
  - [x] when image is too big
  - [x] when image url is not an image
  - [x] when icon or applink is not a url
  - [x] when a new package is detected and no config file is found
  - [x] when adding an invalid config file to an existing package
  - [x] when updated an existing config file using invalid details

### DIFF
--- a/.github/scripts/utils.ts
+++ b/.github/scripts/utils.ts
@@ -1,0 +1,74 @@
+const path = require("path");
+const { exec } = require("child_process");
+const file = require("fs/promises");
+const { promisify } = require("util");
+
+const execAsync = promisify(exec);
+
+async function getNewPackages(): Promise<string[]> {
+  // Get list of all directories in packages/ on main
+  const { stdout: mainDirs } = await execAsync(
+    "git ls-tree -d --name-only main:packages/",
+  );
+  const mainPackagesSet = new Set(
+    mainDirs.split("\n").filter((name: string) => name.trim() !== ""),
+  );
+
+  // Get list of all directories in packages/ in the current HEAD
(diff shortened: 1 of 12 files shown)


mmackz:
Closed the old one since it brought in a bunch of commits from my test branch.

.github/scripts/utils.ts
```diff
@@ -0,0 +1,74 @@
...
+    .filter((pkg: string) => !mainPackagesSet.has(pkg))
+    .map((pkg: string) => path.join('packages', pkg))
+
+  return newPackageDirs
+}
+
+async function getUpdatedPluginDetailsPaths(): Promise<string[]> {
+  // compares the current HEAD with the previous commit to get the updated plugin details
+  const { stdout, stderr } = await execAsync(
+    "git diff --name-only HEAD^ HEAD -- 'packages/*plugin-details.yml'",
```
mmackz, line 36: Fully tested this using multiple scenarios. It is working as intended.

PR #286, opened 2024-03-09
refactor: move plugins into separate directory

mmackz:
Plugins were being mixed with non-plugin packages such as `utils`, `tsconfig` and `registry` in the `packages` directory.

This pr moves all plugin folders into a new directory `packages/plugins`.


Quazia:
This is going to be hard to coordinate without merge issues - we'll want to get any plugin changes upstream before this probably. Don't want regressions to slip through in the migration.

PR #287, opened 2024-03-10
chore(uniswap): bump uniswap-sdk

mmackz:
- Bump and pin uniswap SDK

- Update uniswap tests to use V4 universal router address


packages/uniswap/package.json
```diff
@@ -35,7 +35,7 @@
   "dependencies": {
     "@rabbitholegg/questdk-plugin-utils": "workspace:*",
     "@rabbitholegg/questdk": "workspace:*",
-    "@uniswap/sdk-core": "^4.0.9",
-    "@uniswap/universal-router-sdk": "^2.0.2"
+    "@uniswap/sdk-core": "4.2.0",
+    "@uniswap/universal-router-sdk": "1.7.1"
```
Quazia, line 39: Is this from the weird situation where they like.... updated to an older version?

Quazia (approved):
LGTM

packages/uniswap/package.json
mmackz: CleanShot 2024-03-11 at 14 03 22@2x

I thought it was a bit weird. Not sure what the idea is behind the versioning 🤔

PR #292, opened 2024-03-13
feat(fabric): add support for Fabric mint plugin

mmackz:
## Integration of `fabric` mint action plugin for QuestDK

This PR introduces the integration of the `fabric` mint plugin for QuestDK.

### About The Project

Fabric and the Subscription Token Protocol extends the Non Fungible Token standard with the addition of a temporal dimension, and economic incentives for distribution, adoption, growth, and retention.

- Documentation: https://docs.withfabric.xyz/stp/overview

- Mint Gallery: https://hypersub.withfabric.xyz/

### Implementation Details

The mint plugin targets ERC-721 based subscription NFT mints on the [SubscriptionTokenV1](https://basescan.org/address/0x90e1c94ca5e161acb420e59d8be5041db931c9dd#code) contract. There are 4 functions that mints may be routed through.

- `mint`
- `mintFor`
- `mintWithReferral`
- `mintWithReferralFor`

The `mintFor` methods are used when the NFT is gifted to another address. Referral methods are used when referral metadata is attached to the transaction. Otherwise these all are very similar.

The only parameter on the mint function is `numTokens` which is the same value as `transaction.value` when ETH is used for the mint. 

On `mintFor` there is an extra parameter called `account` which is the recipient of the NFT when the gift option is choosen on the frontend. This parameter will map to the `recipient`.

- only 1 subscription NFT can be minted per transaction, so the `amount` paramenter is ignored.
- `tokenId` is not used for ERC721 mints, so it will be ignored if one is input.

#### Sample Transactions
- [mint](https://basescan.org/tx/0xc0eaf3fa1e32bf3b32ff12e6f5d98b14b46865b62416237b7ed9bc898a5d7d93)
- [mintFor](https://basescan.org/tx/0x5b48379058011a2fd5419b236019b37768ffc2333c03ab097dd6d7fab432c31e)


packages/fabric/README.md
```diff
@@ -0,0 +1,30 @@
...
+### Mint Plugin
+
+The mint plugin targets the mint functions on the [SubscriptionTokenV1](https://basescan.org/address/0x90e1c94ca5e161acb420e59d8be5041db931c9dd#code) contract. There are 4 functions that mints may be routed through.
+
+- `mint`
+- `mintFor`
+- `mintWithReferral`
+- `mintWithReferralFor`
+
+The `mintFor` methods are used when the NFT is gifted to another address. Referral methods are used when referral metadata is attached to the transaction. Otherwise these all are very similar.
```
Quazia, line 19: When we implement the simulation and intent I think this is the function we should hit here. Like the EOA support, that'll be huge when we start using abstracting accounts more.

packages/fabric/src/Fabric.ts
```diff
@@ -0,0 +1,32 @@
+import {
+  type TransactionFilter,
+  type MintActionParams,
+  compressJson,
+} from '@rabbitholegg/questdk'
+import { type Address } from 'viem'
+import { Chains } from '@rabbitholegg/questdk-plugin-utils'
+import { FABRIC_ABI } from './abi'
+
+export const mint = async (
```
Quazia, line 10: lul I love these lil smolbois

Quazia (approved):
EZ :shipit:

PR #293, opened 2024-03-13
feat(aerodrome): add support for Aerodrome swap plugin

mmackz:
## Integration of `aerodrome` swap action plugin for QuestDK

This PR introduces the integration of the `aerodrome` swap plugin for QuestDK.

### About The Project

Aerodrome Finance is a next-generation AMM exclusive to BASE designed to serve as BASE's central liquidity hub, combining a powerful liquidity incentive engine, vote-lock governance model, and friendly user experience. Aerodrome inherits the latest features from Velodrome V2.

- Documentation: https://aerodrome.finance/docs

### Implementation Details

Aerodrome is a fork of Velodrome on BASE, which uses Uniswap V2 style swaps. There are a total of 3 swap methods we will watch for, and their `SupportingFee` counterparts. These use the same parameters as the non-supporting fee functions, so we can expect them to act the same as the functions without supporting fees.

- `ExactETHForTokens`
- `ExactTokensForEth`
- `ExactTokensForTokens`

### Limitations

Like other uniswap V2 forks, AmountOut is not very reliable for some methods (Exact_X_For_X), as it uses amountOutMin, which can be much lower then expected. In extreme cases it can even be zero if slippage is set high enough. This will cause issues when using (==) operator or the (>=) operator.

#### Sample Transactions
- [ExactETHForTokens](https://basescan.org/tx/0x553c5ab29791da1e5b8beb9aaebd517849c06f23356b3188f809062a65472157)
- [ExactTokensForETH](https://basescan.org/tx/0x9af9f272fbf6ecde7e58f3f16f851b51c60cef84e24bde7b77e0d4640835463a)
- [ExactTokensForTokens](https://basescan.org/tx/0xd3a8f0d830d3f43db6ab70ee36c20b34dca587358cabc7d8c083f3d630b8b5f8)


packages/aerodrome/src/Aerodrome.ts
```diff
@@ -0,0 +1,79 @@
...
+
+export const swap = async (
+  _params: SwapActionParams,
+): Promise<TransactionFilter> => {
+  const { chainId, tokenIn, tokenOut, amountIn, amountOut, recipient } = _params
+
+  return compressJson({
+    chainId,
+    value: tokenIn === zeroAddress ? amountIn : undefined,
+    to: AERODROME_ROUTER,
```
Quazia, line 24: We should almost consider a helper functions for this down the line - maybe it takes the router? The rest is pretty bog-standard right?

packages/aerodrome/src/Aerodrome.ts
```diff
@@ -0,0 +1,79 @@
...
+    input: {
+      $or: [
+        {
+          // swapExactETHForTokens
+          $abi: ETH_FOR_TOKENS_FRAGMENTS,
+          amountOutMin: amountOut,
+          routes: {
```
Quazia, line 31: Do we have a helper somewhere for this?

Quazia (approved):
LGTM

packages/aerodrome/src/Aerodrome.ts
mmackz: We have a helper for this, but we have not put it into our utils package. It is used quite often so maybe it would not be a bad idea to add this helper.

I mostly did it this way because I was making a tutorial during the development of this plugin and wanted to show an example of using some of the filter operators. 

I went ahead and switched it to use the helper function.

PR #294, opened 2024-03-13
feat(paragraph): add support for Paragraph mint plugin

mmackz:
## Integration of `paragraph` mint action plugin for QuestDK

This PR introduces the integration of the `paragraph` mint plugin for QuestDK.

### About The Project

Paragraph. xyz is a web3-native publishing platform that helps creators publish blogs and newsletters for virtual communities. The platform aims to help creators reach their audiences on web2 while using web3 tools to create, distribute, and monetize their content.

- Documentation: https://docs.paragraph.xyz/

- Gallery: https://paragraph.xyz/discover/feed/trending/paragraph

### Implementation Details

The mint plugin targets the mint functions on the paragraph [ERC721](https://basescan.org/address/0x3e3255cbe27f34a981a8afa98192e77eb198901a#code) contract. There are 2 functions that mints may be routed through.

- `mint`
- `mintWithReferrer`

Both of these use the same parameters, with the exception of the additional referrer parameter which is not relevant to our plugin.

### Limitations
- only 1 NFT can be minted per transaction, so the `amount` paramenter is ignored.
- tokenId is ignored since these are ERC721
- There is a [createAndMint function](https://basescan.org/tx/0x438ae0d6ed5964ae70a72e34fd71b6c24b252efdd8a56d1c9e628971ae04329d) on the ERC721 Factory that is called when there is a "first mint". Unfortunately this cannot be supported since the contract address for the project is not actually created until the first mint happens. We have had this issue with other mint actions such as zora. You technically need to have the contract address before you can make the boost, so this should be a non-issue.

### Sample Transactions
- [mintWithReferrer](https://basescan.org/tx/0x7946162535d790e488d0168fd89c68d526cfde6c8cf26a01fbcc4e2c78c3040c)


packages/paragraph/README.md
```diff
@@ -0,0 +1,27 @@
...
+- `mint`
+- `mintWithReferrer`
+
+Both of these use the same parameters, with the exception of the additional referrer parameter which is not relevant to our plugin.
+
+
+##### Limitations
+- only 1 NFT can be minted per transaction, so the `amount` paramenter is ignored.
+- tokenId is ignored since these are ERC721
+- There is a [createAndMint function](https://basescan.org/tx/0x438ae0d6ed5964ae70a72e34fd71b6c24b252efdd8a56d1c9e628971ae04329d) on the ERC721 Factory that is called when there is a "first mint". Unfortunately this cannot be supported since the contract address for the project is not actually created until the first mint happens. We have had this issue with other mint actions such as zora. You technically need to have the contract address before you can make the boost, so this should be a non-issue.
```
Quazia, line 24: Reservoir is running into issues around this - they're closer to the NFT problem space so ideally they solve it and we can leverage. Non-trivial problem.

packages/paragraph/src/Paragraph.test.ts
```diff
@@ -0,0 +1,61 @@
...
+import { passingTestCases, failingTestCases } from './test-transactions'
+import { mint } from './Paragraph'
+
+describe('Given the paragraph plugin', () => {
+  describe('When handling the mint action', () => {
+    describe('should return a valid action filter', () => {
+      test('when making a valid mint action', async () => {
+        const filter = await mint({
+          chainId: 1,
+          contractAddress: '0x23d87d8c9704b8bcdbac042b9a59a142f0f10298',
```
Quazia, line 12: no dead beef 😆

Quazia (approved):
LGTM

PR #295, opened 2024-03-13
getFees - break fees up into action and project

benjaminykim:
- previously Sound, Mirror, and Zora were only returning the total fees associated with a project
- now, we break up fees into an object `{ projectFees: BigInt, actionFees: BigInt }` in a non-destructive way to ensure backwards compatibility
- tests written to match pattern for `getProjectFees`
- documentation added where needed (referenced Soundxyz sdk)


packages/soundxyz/src/Soundxyz.ts
```diff
@@ -163,6 +163,38 @@ export const getProjectFees = async (
...
+
+  const totalPriceAndFees = (await client.readContract({
+    address: SUPERMINTER_V2,
+    abi: TOTAL_PRICE_AND_FEES_ABI,
+    functionName: 'totalPriceAndFees',
+    args: [contractAddress, mintInfo.tier, mintInfo.scheduleNum, 1, false], // assume quantity is 1 and hasValidAffiliate is false
+  })) as TotalPriceAndFees
+
+  return {
+    actionFee: totalPriceAndFees.subTotal,
```
jonathandiep, line 193: Should this return `totalPricesAndFees.total`?

packages/mirror/src/Mirror.ts
```diff
@@ -131,6 +131,46 @@ export const getProjectFees = async (
   return platformFee + nftPrice
 }
 
+export const getFees = async (
```
mmackz, line 134: No issues, this looks good to me. 

Just want to point out that you can combine the multiple reads using viems multicall.

https://viem.sh/docs/contract/multicall#multicall

mmackz:
You will need to generate a changeset.

packages/soundxyz/src/Soundxyz.test.ts
```diff
@@ -161,3 +161,23 @@ describe('getProjectFees', () => {
     expect(fee).toEqual(BigInt('777000000000000'))
   })
 })
+describe('getFees', () => {
```
Quazia, line 164: These tests arn't super meaningful but this is in line with the standard of tests for this function so NBD, I can clean these up at the same time as those.

Quazia (approved):
This looks good - needs changeset like @MackeDev said. I don't like doubling out the logic in a second function when we could call `getFees` from `getProjectFees` to avoid the duplicated code, and we could probably even just recycle getFees in the registry [getProjectFees](https://github.com/rabbitholegg/questdk-plugins/pull/295/files/eb655f489b46078eb821c79fe34fed904620d60a#diff-5e23fa7ab384750c59928974f281315a01f49d7e8751645b5a253fae32cf7890R141). I also wouldn't worry too much about being non-destructive _now_ since we can always push an update.

I like the action fee vs project fee that's a good solution.

Quazia:
Also please try to use atomic [conventional commits](https://github.com/rabbitholegg/questdk-plugins?tab=readme-ov-file#commit-standards) as much as possible in the public repos for maximum external readability.

packages/soundxyz/src/Soundxyz.ts
benjaminykim: /*
  https://github.com/soundxyz/sound-protocol/blob/f91f87739c1626b8d2d4e9f02ba573671c627698/contracts/modules/interfaces/ISuperMinterV2.sol#L111C1-L126C6
  struct TotalPriceAndFees {
    uint256 total: (`subTotal + platformTxFlatFee + artistReward + affiliateReward + platformReward`).
    uint256 subTotal: The total price before any additive fees.
    uint256 unitPrice: The price per token.
    uint256 finalArtistFee: The final artist fee (inclusive of `finalArtistReward`).
    uint256 finalAffiliateFee: The total affiliate fee (inclusive of `finalAffiliateReward`).
    uint256 finalPlatformFee: The final platform fee (inclusive of `finalPlatformReward`, `perTxFlat`, sum of `perMintBPS`).
  }
*/

checking the docs, subTotal is basically unitPrice * number of units... I think?

Quazia:
Thank you

benjaminykim:
> Also please try to use atomic [conventional commits](https://github.com/rabbitholegg/questdk-plugins?tab=readme-ov-file#commit-standards) as much as possible in the public repos for maximum external readability.

ok I `git rebased` to clean up my commit history, can @Quazia take a look?

Quazia (approved):

PR #296, opened 2024-03-14
Feat(rollup): changes our build to a three step process using tsc, babel and rollup

Quazia:
This package is still way bigger than it should be but this helps makes the modules more portability and gives us a path to slimming down the build through dedupes and exclusions. This at least gives us a foundation to start that work.

Unfortunately I don't have any metrics around build time or on the compilation within RH front-end but anecdotally it's integrating MUCH smoother and faster now, I would guess from the ability to do more precise tree shaking.

**This PR focuses on:**
1. Breaking out the build process into three discrete steps
2. Starting to [migrate away from Rome](https://github.com/rome/tools?tab=readme-ov-file#rome-wont-be-maintained-anymore-by-the-same-people-that-maintained-it-so-far-biome-will-provide-new-features-and-fixes) as it's no longer maintained

**Our new build flow goes as follows:**
1. Verify types and build type definitions.
2. Transpile into unbundled CJS and ESM using Babel
3. Bundle into ESM and UMD, including minifyed versions of each using rollup

This has the benefits of type checks and safety since we do this _first_ but the speed of unchecked transpilation and bundling since babel doesn't type check leading to faster build times that shake better. Injecting Babel into our build step also helps us resolve some of the issues we run into from importing less portable SDKs which is a requirement of the plugin system.

I've also added a helper for copying files out to every directory as sometimes it's cleaner to delete all and replace them.

There are still some tasks that should be addressed with the rollup, mainly the node polyfills, but I'm running into issues around `TextDecoder` which is not available as a polyfill from [rollup-plugin-polyfill-node](https://www.npmjs.com/package/rollup-plugin-polyfill-node). I am filling the node utils for most packages, just not for ones that use Axios since that relies on `TextDecoder`. Also, `TextDecoder` should be available in the web environment I just have found a good way to selectively fill the node utils. For now just understand that if we _do_ bring this into a web browser environment we're going to need to fill some node utils which should be doable with webpack. I'll keep looking into this more.


mmackz:
Is a changeset needed?

mmackz (approved):
looks good.

jonathandiep (approved):
We have tested running this locally so LGTM!!

May need to fix conflicts before merging

package.json
```diff
@@ -1,10 +1,11 @@
 {
   "name": "questdk-plugins",
+  "type": "module",
   "scripts": {
     "create": "pnpm build && pnpm run -r --filter=@rabbitholegg/create-plugin create",
-    "build": "turbo run build --filter=!@rabbitholegg/questdk-plugin-project",
+    "build": "NODE_OPTIONS='--max-old-space-size=6096' turbo run build --filter=!@rabbitholegg/questdk-plugin-project,!@rabbitholegg/questdk-plugin-connext",
```
mmackz, line 6: Why excluding connext from the build step? Im a bit curious 🧐

Quazia: This doesn't actually work, I've gotta remove this line but for some reason connect is **eating** RAM during the build step

PR #306, opened 2024-03-18
feat(oku): add pool options for a given token

abs3ntdev:
This PR adds the ability to get a list of all Uniswap pools containing a given token using the Oku API.

This will be used for tracking limit order quests on Oku.


mmackz:
This looks great guys. Lets get this into staging.

One thing I noticed is that it is not possible to use raw ETH for limit orders. (also not sure if zeroAddress will work on the `getPools` function?) You may want to filter ETH out of the token list if you wish to use the same one. 

https://github.com/rabbitholegg/questdk-plugins/blob/31b1440628fa731fb5577b2c1719c1d166d4d260/packages/okutrade/src/OkuTrade.ts#L92-L94

You can add the actionType to `getSupportedTokens` to give different token lists for each action.

```typescript
export const getSupportedTokenAddresses = async (
  _chainId: number,
  actionType?: ActionType,
): Promise<Address[]> => {
  if (actionType === 'options') {
    return CHAIN_TO_TOKENS[_chainId]?.filter((token) => token !== zeroAddress) ?? []
  }
  return CHAIN_TO_TOKENS[_chainId] ?? []
  ```

**Comment by abs3ntdev** [comment]
I added the filter for zeroAddress and adjusted the test and ensure it is not included in the supported token list

**Review by mmackz — APPROVED** [review]

File: packages/okutrade/src/utils.ts [hunk 1]
```diff
@@ -56,6 +88,15 @@ const chainToWETH: Record<number, Address> = {
   [Chains.BASE]: '0x4200000000000000000000000000000000000006',
 }
 
+const chainIdToName: Record<number, string> = {
+  1: 'ethereum',
```
mmackz, line 92: one small nit, but could you use the Chains enum here like used in `chainToWeth`

abs3ntdev: done

PR #307, opened 2024-03-19
feat: add JokeRace proposals

seanmc9:
Left TODOs:
- [x] add `contractAddress` param to everything
- [ ] fill out test-transactions + correct params in JokeRace.test.ts
- [ ] jokerace README + CHANGESET


Quazia (commented):
This looks good! One thing I want to flag is that Boost Actions as defined in the utils package are Types that are pretty heavily embedded in both our front and backend, so it takes a decent amount to support novel Action Types. Ideally new plug-ins integrate with existing Actions even if it takes some adjustment for it to line up as expected. 

Can you explain a little more about the proposal parameter on the propose and proposeWithoutProof functions?

If we were to create a proposal action it'd need to be a single Action Type and it would need to work for all voting systems that accept proposals (needs to be re-usable across protocols) as we don't really support custom Action types for any single protocol.

If it helps we populate [this flow
](https://manager-staging.boost.xyz/boosts/create) from the plugins in this repo.

Let me know if you have any questions and I'm excited to see what you do! Jokerace is a cool project and excited to collab!

seanmc9:
@Quazia aah gotcha, yeah that all makes sense - that was my biggest question re: if y'all would be open to supporting bespoke actions for us, because unfortunately yeah we don't have a super standard `propose()` function on JokeRace contests and diverge pretty solidly from pre-existent standards like Governor or things like that.

wrt your question re the proposal parameter:
- for reference - [here](https://github.com/jk-labs-inc/jokerace/blob/44da767a036a9c73f0988babc52bc65d38ffbd4e/packages/forge/src/governance/Governor.sol#L351) is the `propose()` function in our Solidity codebase and [this](https://github.com/jk-labs-inc/jokerace/blob/44da767a036a9c73f0988babc52bc65d38ffbd4e/packages/forge/src/governance/Governor.sol#L58) is the ProposalCore struct (what that proposal param is). 
- tldr on the parameter is that we package everything we need into it particularly to make metadata as modular as possible/we include all of the possible metadatas in every proposal and just set them all to 0 if the user isn't using them - allows us to add metadata options in the future without having to overload the `propose()` function a ton like other implementation options might require us to.

and thank you, I really appreciate that! excited to collab as well!

Quazia:
@seanmc9 

Ok so realistically we would probably want to pass in _those_ as params vs the structured proposal. The only elements we need to parameterize are the ones you'd want to control for the boost. So for example for a Swap we want to be able to require a certain token or amount is swapped to be considered a complete boost i.e. swap 500 $DEGEN. For the proposal case what params would you want to be able to run boosts against? As you have it now the only quests you could run would be for an EXACT proposal which probably wouldn't be too useful.

Also we talked internally, it would be better to start with a Vote action since we support that now. We can support proposals at some point just not sure when. Another option would be to shoehorn proposal action into vote until we have first class support and then split vote out at that point.

Quazia:
[Here's an example](https://github.com/rabbitholegg/questdk-plugins/blob/main/packages/gmx/src/GMX.ts) of a more complicated plugin to show you the sort of options we have to parse out the proposal param.

seanmc9:
@Quazia gotcha yeah that makes sense, yeah the context I needed was that bespoke actions aren't really an option, so it's really about how can we put the relevant data to read JokeRace `propose()` transactions into the general action models that are already set up.

re what I was thinking the boost would be on - each JokeRace contest is its own contract, and each contest has multiple proposals in it, so what I was thinking you'd do the boost on is whether someone has proposed to a given contest (an example here could be [this Hats contest](https://www.jokerace.io/contest/optimism/0x5c11016ee4f8ad4ea2ab8b1b366f32d30d48a031) that anyone can submit to).

that makes sense with doing voting first (the boost being if you voted on a proposal in a contest) though - that's probably the move first and then when/if proposals become a supported generic action seeing how we can make JokeRace proposals work within that model makes sense!

seanmc9:
So with voting these are our function signatures:
- `castVote(uint256 proposalId, uint8 support, uint256 totalVotes, uint256 numVotes, bytes32[] calldata proof)`
- `castVoteWithoutProof(uint256 proposalId, uint8 support, uint256 numVotes)`

which seem like they're pretty compatible with the general vote action model y'all have!

maybe the move is here to close this PR and make a new one for voting on JokeRace as the first integration? curious on your thinking here and/or if I'm understanding things correctly

Quazia:
@seanmc9 yeah that all sounds right except for some of the details on the proposal for proposals 😅 I'll have to look at your contract more and think about the best direction for a general proposal action to have a strong opinion there but the path forward with vote seems solid 🙏🙌

seanmc9:
Sounds good! I'll go ahead and close this PR and open a new one for voting!

seanmc9:
@Quazia, I created the new PR in #308

PR #308, opened 2024-03-19
feat: add jokerace voting plugin

seanmc9:
TODOs:
- [x] figure out how to shoehorn `contractAddress` and `numVotes` into `packages/jokerace/src/JokeRace.ts`
- [x] fill out test files
- [x] update README


seanmc9:
@Quazia so I've fiddled with things a bit and I think that if it could be possible to add a `contractAddress` parameter and a `numVotes` param (both of which should be pretty standard I believe - the contract that the vote txn is going to and the number of votes being cast) to the `VoteActionParams`, then we could make our `castVoteWithoutProof()` function work with y'all's!

seanmc9:
looks like the only other plugin that uses `VoteActionParams` is the Llama one, so hopefully wouldn't be too bad to update their test files to accommodate this schema update either.

Quazia:
@seanmc9 Is it possible to use the [project](https://github.com/rabbitholegg/questdk-plugins/pull/308/files#diff-d9c3b5f3c46529e88081e868944bc351ded1a4b66e49ebfa6adb72fb5c65da19R83) param instead of the add contract address? I think the `numVotes` works. Maybe `weight` would be another name option.

seanmc9:
@Quazia Oh yes, absolutely, great call! Didn't know if `project` was a reserved param in y'all's system since it's so prevalant, but if that could serve as the contract address that'd be great! And yeah, `weight` works too name-wise!

seanmc9:
@Quazia updated! does this structure look good to flesh out the tests for?

Quazia:
@seanmc9 Yes it looks great! Sorry for the delay.

seanmc9:
@Quazia no worries!

seanmc9:
@Quazia Alrighty, just updated the rest of the files! Does everything look good here/ready to merge?

mmackz:
Can you bring in the latest changes from our main branch and regen the lockfile. There is a few conflicts.

mmackz:
You will also need to generate a changeset. Run the changeset command from root and follow the instructions.

`pnpm changeset`

seanmc9:
@mmackz done!

seanmc9:
@Quazia @mmackz just pinging here again to see if there's anything else that's needed!

PR #312, opened 2024-03-20
Feat(sound): fix mint fees

Quazia:
So turns out sometimes if there's only 1 tier it's actually tier 1 and not tier 0.

This handles a bunch of weird edge cases for sound mint fees and zora functions when tokenId is undefined.

### DIFF
--- a/packages/zora/src/Zora.ts
+++ b/packages/zora/src/Zora.ts
@@ -151,32 +151,42 @@ export const simulateMint = async (
       transport: http(),
     })
   const from = account ?? DEFAULT_ACCOUNT
+  let _tokenId = tokenId
+  if (tokenId === null || tokenId === undefined) {
+    const nextTokenId = (await _client.readContract({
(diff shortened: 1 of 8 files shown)


packages/zora/src/Zora.ts
```diff
@@ -151,58 +151,68 @@ export const simulateMint = async (
...
+      functionName: 'nextTokenId',
+    })) as BigInt
 
-  if (tokenId !== null && tokenId !== undefined) {
+    _tokenId = Number(nextTokenId) - 1
+  }
+
+  try {
+    const mintArgs = [
+      FIXED_PRICE_SALE_STRATS[mint.chainId],
```
jonathandiep, line 167: @mmackz found out that there could be mutliple sale strat addresses... we don't need to fix in this PR, but will want to address

jonathandiep (approved):
Played around with this locally and lgtm! Will go through publish flow

PR #314, opened 2024-03-20
fix: Use zora sdk to grab the fixed price sale strategy

jonathandiep:
Sometimes mints would fail because it is using the wrong fixed price sale strategy

### DIFF
--- a/packages/zora/src/Zora.ts
+++ b/packages/zora/src/Zora.ts
@@ -109,12 +109,16 @@ export const getMintIntent = async (
   let data
 
   let fixedPriceSaleStratAddress = FIXED_PRICE_SALE_STRATS[chainId]
-  
+
   try {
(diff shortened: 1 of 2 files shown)


packages/zora/src/Zora.ts
```diff
@@ -105,11 +105,25 @@ export const mint = async (
...
   mint: MintIntentParams,
 ): Promise<TransactionRequest> => {
-  const { contractAddress, tokenId, amount, recipient } = mint
+  const { chainId, contractAddress, tokenId, amount, recipient } = mint
   let data
+
+  let fixedPriceSaleStratAddress = FIXED_PRICE_SALE_STRATS[chainId]
+
+  try {
+    console.log({ chainId, contractAddress, tokenId })
```
jonathandiep, line 114: remove log

PR #322, opened 2024-03-25
Mmackz/fix_zora_simulate

mmackz:
The NFT used for the `simulateMint` function has a time-limited mint which will expire next week. After this mint expires, the simulations will fail.

This pr replaces the current test with one that uses an NFT that will remain open indefinately.


Quazia (approved):
LGTM :shipit:

PR #323, opened 2024-03-26
feat(utils): add sepolia chain to viem chain helper

mmackz:
adds support for sepolia chain in the `chainIdToViemChain` helper utility.


Quazia (approved):
:shipit:

PR #325, opened 2024-03-26
feat(fabric): add projectFees/simulation functions

mmackz:
This PR adds the functionality needed to be supported by boost inbox.

- getFees
- getProjectFees
- getMintIntents
- simulateMint

Only NFTs which use ETH as the currency will be able to pass simulation. (The idea being users will not be able to deploy if the mint is not ETH denominated) 

At the moment we do not have support for ERC20 denominated mints because of how our cross-chain and embedded mints work. The majority of mints are denominated in ETH, so this should not be too big of an issue to not support ERC for now.

Needs #323 to be merged before this will pass tests.


Quazia:
Should we close this for now? Until we want to support ERC20 fees at least.

Quazia (approved):
LGTM 🙏

PR #326, opened 2024-03-26
feat(paragraph): add projectFees/simulation functions

mmackz:
This PR adds the functionality needed to be supported by boost inbox.

- getFees
- getProjectFees
- getMintIntents
- simulateMint

This project is compatible with Polygon, so some mints will use native MATIC as payment for the mint. We will need to take this into consideration when we are converting to USD values on the frontend.


Quazia (approved):
LGTM - front end is setup to handle other native tokens for the USD conversion so we should be good to go.

PR #327, opened 2024-03-27
feat(kwenta): add Kwenta options support in Base

rontero83:


packages/kwenta/src/Kwenta.ts
```diff
@@ -0,0 +1,56 @@
...
+  // We always want to return a compressed JSON object which we'll transform into a TransactionFilter
+  const { chainId, recipient, amount } = _params
+
+  return compressJson({
+    chainId,
+    to: SMART_MARGIN_V3,
+    from: recipient,
+    input: {
+      $abiAbstract: SMART_MARGIN_V3_COMMIT_ORDER_ABI,
+      _sizeDelta: amount,
```
mmackz, line 43: We will also need to add a way to determine the tokenAddress, and orderType. This data should hopefully be encoded in the transaction details.

tokenAddress for example will likely be able to be determined from the _perpsMarketId

You will need to find out the mapping between marketId and tokens. marketId of 100 seems to be ETH-USD for example.

mmackz: 100 = ETH-USD
200 = BTC-USD (you can use wBTC for this)

rontero83: @mmackz You mean that only these two pairs should be used, don't you?

mmackz: These are the only available pairs.

CleanShot 2024-03-29 at 08 06 02@2x

mmackz (approved):

PR #329, opened 2024-03-29
feat(sound): add support for legacy contract in simulations

mmackz:
This PR adds support for the legacy `SUPERMINTER` contract to pass simulations and calculate fees.

CleanShot 2024-03-28 at 17 25 14@2x

### DIFF
--- a/packages/soundxyz/src/Soundxyz.ts
+++ b/packages/soundxyz/src/Soundxyz.ts
@@ -204,12 +204,7 @@ export const getFees = async (
       address: SUPERMINTER,
       abi: TOTAL_PRICE_AND_FEES_V1_ABI,
       functionName: 'totalPriceAndFees',
-      args: [
-        contractAddress,
-        tier,
-        BigInt(0),
-        quantity,
-      ],
+      args: [contractAddress, tier, BigInt(0), quantity],
     })) as TotalPriceAndFees
 
     return {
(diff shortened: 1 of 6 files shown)


Quazia (approved):
LGTM - only callout is that the `mintTo` function looks the same on both ABIs. I see that the fee view call is different but maybe we use the same ABI for both and recycle the `mintTo` function between the two? I don't know, doesn't really seem like it matters one way or another but it feels wrong to have the same function duplicated. Not in any way blocking though. Thanks for the fix.

Quazia:
@mmackz Can you walk me through the issue with `getDefaultMintTier` is it just needing to fallback to the other address? It does look like it implements `nextScheduleNum` and returns what we'd expect. The example you're using has a tier 1 and a pre-release schedule so we might need to test with the GA only case and the no presale case?

packages/soundxyz/src/Soundxyz.ts
```diff
@@ -226,13 +227,26 @@ export const getDefaultMintTier = async (
...
+
+  // check to see if contract uses legacy SUPERMINTER
+  // will return an array of mintInfo or an empty array
+  const mintInfoListResult = (await _client.readContract({
     address: SUPERMINTER,
+    abi: MINT_INFO_LIST_ABI,
+    functionName: 'mintInfoList',
+    args: [contractAddress],
+  })) as unknown[]
+  const isLegacy = mintInfoListResult.length
```
Quazia, line 239: That's slick

Quazia (approved):
LGTM

PR #339, opened 2024-04-03
feat(builder): add blast chain support for builder

mmackz:
- bumps viem version (2.7.9 -> 2.9.9)

- add blast chain to builders viem client. This allows it to look up transaction hashes on blast network.

- use Chains enum


Quazia (approved):
LGTM

PR #344, opened 2024-04-04
feat(sushi): add support for BLAST chain

mmackz:
- add list of tokens for BLAST
- updates route processor addresses to V4
- add blast to list of supported networks
- add blast specific test


Quazia (approved):
LGTM - has this plugin been broken for the V4 processors?

mmackz:
> LGTM - has this plugin been broken for the V4 processors?

It seems that way. I dont see any activity on them for the past month.


Pulled them from this page.

https://github.com/sushiswap/sushiswap/blob/master/protocols/route-processor/deployments/base/RouteProcessor4.json

mmackz:
It has been broken since about late February 😬

They do have an SDK, but if I remember correctly it had some inconsistencies. (contracts didn't match what I was seeing being used by the frontend)

Maybe it was just bad timing, or I was using the wrong SDK version, but might be worth trying to pull the contracts off the SDK again if we run into more issues.

Quazia:
As long as we didn't have user reports NBD I guess? We would have found it if we did.

PR #345, opened 2024-04-04
feat(synapse): add blast support to synapse plugin

mmackz:
- add blast support
- uses default token list for synapse bridge (USDB, ETH)
- add blast specific test
- add additional testing coverage

No CCTP contract is on BLAST.


Quazia (approved):
LGTM

mmackz:
Made some changes to use the chains enum from the utils package.

PR #348, opened 2024-04-09
feat(orbit): add support for Orbit stake plugin

mmackz:
## Integration of `orbit` stake action plugin for QuestDK

This PR introduces the integration of the `orbit` swap plugin for QuestDK. 

### About The Project

Orbit is a decentralized liquidity protocol that facilitates the lending and borrowing of Blast assets. Orbit's innovation is to make use of Blast's native yield to provide a better lending / borrowing experience.

- **Documentation** - https://book.orbitlending.io/

### Implementation Details

The stake plugin targets the VE_ORBIT token contract which allows users to stake their ORBIT tokens in exchange for yield, and blast points. There is only one compatible token (ORBIT) and you can lock your stake for a higher rate of reward emissions. Staking times can range from 1 day to 4 years. The lock time is shown as a timestamp (in seconds) which is the time you will be allowed to withdraw your tokens.

The contract function will will be filtering for is:

- `create_lock`

### Sample Transactions

- [create_lock](https://blastscan.io/tx/0x9e4dadb4b8be32aa979367940a0316a0270aea5988853d26124568ac0a092fd8)


packages/orbit/src/utils.ts
```diff
@@ -0,0 +1,16 @@
...
+  if (amount === undefined) return undefined
+
+  // in seconds
+  const now = Math.floor(new Date().getTime() / 1000)
+
+  if (typeof amount === 'object') {
+    const [operator, value] = Object.entries(amount)[0]
+    return { [operator]: BigInt(value) + BigInt(now) }
+  }
+  return { $gte: BigInt(amount) + BigInt(now) }
```
Quazia, line 15: Can you talk me through this? It looks like we're basically requiring that the unlock time is 1 second in the future for every 1 wei of value?

Quazia: Nevermind - I understand now.

I would change `amount: Amount` to `duration: Amount` or even `duration: Duration`. Also I think we should have that type somewhere but if we don't let's export it from utils vs define it here. That's a pretty useful/re-usable one.

Quazia (approved):
LGTM - some nits around naming of `amount` vs `duration`

packages/orbit/src/utils.ts
mmackz: duration will be passed in as the amount of time in seconds.

The `_unlock_time` is a timestamp which denotes the exact time that the stake will become unlocked.

The logic here is that The current time (date.now()) + the duration in seconds should meet or exceed the `_unlock_time` timestamp shown in the transaction to pass (if the operator is $gte)

PR #350, opened 2024-04-09
feat(okutrade): add blast support

mmackz:
- Adds blast support for both swap and options actions
- Removes all chains except for blast for the swap plugin. Blast is the only network we can support due to the use of unsupported routers on other networks.


packages/okutrade/src/OkuTrade.ts
```diff
@@ -86,7 +87,10 @@ export const options = async (
   })
 }
 
-export const getSupportedChainIds = async () => {
+export const getSupportedChainIds = async (actionType?: ActionType) => {
```
Quazia, line 90: I remember we added this support for another plug-in but we'll want to double check this is still supported on the front-end and behaves nicely/as expected.

Quazia (approved):
LGTM

packages/okutrade/src/OkuTrade.ts
mmackz: It was for the paraswap stake action. It is still working correctly showing a different list of networks on swap/stake.

PR #351, opened 2024-04-10
fix(zora): fix issue with misconfigurations passing simulation

mmackz:
Fixes mint simulations in Zora plugin. 

This will now check to ensure the address is an actual contract by checking for the presence of bytecode.

### DIFF
--- a/packages/zora/src/Zora.ts
+++ b/packages/zora/src/Zora.ts
@@ -1,40 +1,40 @@
 import {
-  type TransactionFilter,
-  type MintActionParams,
-  compressJson,
-} from '@rabbitholegg/questdk'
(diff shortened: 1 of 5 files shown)


packages/zora/src/Zora.ts
```diff
@@ -175,6 +175,12 @@ export const simulateMint = async (
     _tokenId = Number(nextTokenId) - 1
   }
 
+  // check to see if the address is a contract
+  const bytecode = await _client.getBytecode({ address: contractAddress })
+  if (!bytecode) {
+    throw new Error('Address is not a contract')
+  }
```
jonathandiep, line 182: this is a neat way to check for a contract!

jonathandiep (approved):

Quazia (approved):

PR #358, opened 2024-04-25
fix(across): fix verification issue

mmackz:
Fixes an issues where doing any/any quests will only look for the spokepool and not the verifier contract. Native ETH txs would not be detected previously.

- Adds additional tests


Quazia (approved):
LGTM - good thing they handle value in amount!

PR #362, opened 2024-04-30
feat(camelot): add support for native aggregator

mmackz:
- updates camelot plugin to support new native aggregation contracts

https://arbiscan.io/address/0x99D4e80DB0C023EFF8D25d8155E0dCFb5aDDeC5E#code


packages/camelot/src/Camelot.test.ts
```diff
@@ -1,210 +1,40 @@
...
-                      $gte: '500000000000000',
-                    },
-                    partner: '0x353D2d14Bb674892910685520Ac040f560CcBC06',
-                  },
-                },
-              ],
-            },
-          ],
-        },
+        chainId: 1,
```
jonathandiep, line 12: Don't think mainnet is the right network for Camelot

jonathandiep (approved):
lgtm, but left comment about chain id

packages/camelot/src/Camelot.test.ts
mmackz: This test checks that the format of the action filter is correct based on the inputs. 

Its a standard test I started including in all new plugins.

It checks:

  - That the filter is an object
  - It has a chainId property (or sourceChainId if bridge)
  - The chainId matches the input chainId
  - It has a "to" property of type Address or an object with an $or property as the only key
  - It has a valid abi operator

PR #363, opened 2024-04-30
Feat(neynar:follow): add neynar validation for the follow action

Quazia:


packages/neynar/src/Neynar.ts
```diff
@@ -0,0 +1,124 @@
...
+          address: actor,
+          questId: questId,
+          taskId: taskId,
+        }
+      } else {
+        return null
+      }
+    }
+    default:
+      return null
```
mikesizz, line 45: Throw here instead to prevent false positives.

mikesizz (approved):

mikesizz (approved):

PR #378, opened 2024-05-09
feat: add support for multiple standard exit addresses

Quazia:
When a transaction goes through a contract on it's way to the final destination contract (multi-call, bridge relay, account abstraction, etc) we need to be monitoring the address in the filter. This adds a helper that builds that $or statement and seeds it with the layer zero addresses.


sammccord (approved):
looks good to me 👍🏼

PR #385, opened 2024-05-14
[WIP] chore(tooling): use vite for everything, remove unnecessary tsconfig package

sammccord:
discussion point: do we want to bundle any external dependencies like third-party sdks? or leave them external?

- remove all babel/rollup configs
- remove tsconfig package + move shared config to root, it's an unnecessary extra dep resolution for downstream consumers
- use same vite config / exports structure for all packages
- externalize @rabbithole deps so we're not bundling thousands of kb of our own code when downstream bundlers can handle those resolutions themselves


sammccord:
> When you say "externalize @rabbitHole deps so we're not bundling thousands of kb of our own code when downstream bundlers can handle those resolutions themselves" you mean refining the include in the root ts config?

I mean, vite wants to bundle everything, including our workspace packages into a bundle with no external dependencies. This can lead to larger modules, but since we're in library mode it's safe to let our own dependencies / third party sdk's be external - ie `import`'d and `require`'d because downstream consumers will be bundling them themselves and can decide how they want to resolve those deps

PR #387, opened 2024-05-15
feat(Zora): add `create` action to Zora plugin

mmackz:
Creates a new `create` action for the Zora plugin. This action filter will detect any transaction which use the `createContract` function on the `ZORA_1155_FACTORY` contract. The is no additional parameters besides the chainId and contract address.

To-do

- [x] Create types for `create` action type in questdk.


Quazia (approved):
LG - how do you wanna handle the Create Action params? This is failing cause of that seems

PR #390, opened 2024-05-16
feat(utils): add types for create action type

mmackz:
- adds the required types and schema for the `create` action type


packages/utils/src/types/actions.ts
```diff
@@ -244,6 +246,23 @@ export const FollowActionFormSchema = z.object({
...
 
+/*
+CREATE
+*/
+export type CreateActionParams = {
+  chainId: number
+  contractAddress?: Address
+}
+
+export const CreateActionFormSchema = z.object({})
```
mmackz, line 257: Using `null` for the FormSchema throws a bunch of errors 🫠

jimobrien (approved):
🔥

mikesizz (approved):

PR #391, opened 2024-05-17
[DONT MERGE]feat(hai): add support for Hai `stake` plugin

mmackz:
## Integration of `hai` stake action plugin for QuestDK

This PR introduces the integration of the `hai` stake plugin for QuestDK.

### About The Project

Hai is a decentralized platform that operates similarly to DAI, utilizing a Collateralized Debt Position (CDP) system. Users can deposit various types of collateral to mint the stablecoin called HAI.

- Documentation: https://docs.letsgethai.com/detailed/intro/hai.html

## Implementation 
The HAI plugin utilizes the Stake Action Type and looks for deposits of colleteral made using the `openLockTokenCollateralAndGenerateDebt` function. There is a 150 HAI minimum that needs to be minted, but the amount of collateral deposited will exceed this.

There is 4 tokens currently compatible with HAI as collateral. WETH, STETH, SNX, and OP.

This plugin only covers the creation of new vaults. Minting HAI from an existing vault will not be considered, as there is no way to get the context of the collateral token from the tx details.

Since each vault using a different address, we are not able to match the `to` parameter with the contractAddress as we normally do. Instead the implementation contract address is found of the `_target` parameter in the tx data. 

## Sample Transactions
- [openLockTokenCollateralAndGenerateDebt](https://optimistic.etherscan.io/tx/0x2c07c670182229bc1201087c9883ed67c7d46dcbd0f65c2b4d701949705a4581)

---

https://linear.app/rh-app/issue/BD-74/plugin-build-hai

---


packages/hai/src/Hai.ts
```diff
@@ -0,0 +1,54 @@
...
+import { type Address } from 'viem'
+import { Chains } from '@rabbitholegg/questdk-plugin-utils'
+
+export const stake = async (
+  stake: StakeActionParams,
+): Promise<TransactionFilter> => {
+  const { chainId, tokenOne: token, amountOne: amount } = stake
+
+  return compressJson({
+    chainId,
```
mmackz, line 23: Im not sure if this will work as intended. 

Since we are not using the `to` parameter, is there any way someone could exploit this?

Each vault has its own contract.

I set the _target to the `BASIC_ACTIONS_CONTRACT` which is the implementation contract for the vault.


lmk if you want to sync and ill go through this in more detail.

mmackz: GPT thinks it is okay 

>Yes, your approach to filter transactions based on the execute function, with a specific _target address (0x5c03a6562ddd223275b644079fadbc14e4485c52) and specific calldata in _data for the openVaultAndGenerateDebt function, will generally work as intended.

>For a transaction to be successful, the calldata must be valid and the _target contract must execute the function correctly. If the calldata is incorrect or malformed, the transaction will fail.

mmackz:
```
function execute(address _target, bytes _data)
        public
        auth
        note
        payable
        returns (bytes32 response)
    {
        require(_target != 0x0);

        // call contract in current context
        assembly {
            let succeeded := delegatecall(sub(gas, 5000), _target, add(_data, 0x20), mload(_data), 0, 32)
            response := mload(0)      // load delegatecall output
            switch iszero(succeeded)
            case 1 {
                // throw if delegatecall failed
                revert(0, 0)
            }
        }
    }

jimobrien (approved):

packages/hai/src/Hai.ts
jimobrien: I do think this is exploitable - one could deploy a contract that implements the same function interface, pass the right params, and claim rewards without actually completing the action

PR #393, opened 2024-05-17
fix(superbridge): fix issue with invalid transactions passing

mmackz:
- Fixes an issue where bridging ETH was able to pass a boost for a specific token.

It now removes any ETH based input filters if tokenAddress is an ERC20.

CleanShot 2024-05-17 at 12 00 33@2x

https://linear.app/rh-app/issue/BOOST-4002/possible-bug-in-superbridge-plugin

### DIFF
--- a/packages/superbridge/src/Superbridge.ts
+++ b/packages/superbridge/src/Superbridge.ts
@@ -21,11 +21,117 @@ import {
 import { Chains } from '@rabbitholegg/questdk-plugin-utils'
 import { type Address, zeroAddress } from 'viem'
 
+function getInputsFromEthereum(params: BridgeActionParams) {
+  const { tokenAddress, amount, recipient } = params
+  const ercInputs = [
+    {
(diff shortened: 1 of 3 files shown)


packages/superbridge/src/Superbridge.ts
```diff
@@ -21,11 +21,117 @@ import {
...
+      _amount: amount,
+      _to: recipient,
+    },
+  ]
+  const ethInputs = [
+    {
+      $abi: [BRIDGE_ETH_FRAGMENT, DEPOSIT_ETH_FRAGMENT],
+    },
+    {
+      $abi: [BRIDGE_ETH_TO_FRAGMENT, DEPOSIT_ETH_TO_FRAGMENT],
```
mmackz, line 55: All these ABI functions were combined in an $or before. Now it splits them based on the `tokenAddress`.

Now these will only be accesses if tokenAddress is ETH, or tokenAddress is 'Any'

sammccord (approved):

PR #405, opened 2024-05-21
Enable State Override

jonathandiep:
Increases the balance when running `simulateContract` to prevent the low balance `ContractFunctionExecutionError`

```
ContractFunctionExecutionError: The total cost (gas * gas fee + value) of executing this transaction exceeds the balance of the account.
```

### DIFF
--- a/packages/zora/src/Zora.ts
+++ b/packages/zora/src/Zora.ts
@@ -100,9 +100,9 @@ export const mint = async (
 
   const mintContracts = universalMinter
     ? ([
-      contractAddress.toLowerCase(),
-      universalMinter.toLowerCase(),
-    ] as Address[])
+        contractAddress.toLowerCase(),
+        universalMinter.toLowerCase(),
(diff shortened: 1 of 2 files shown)


packages/zora/src/Zora.ts
```diff
@@ -282,11 +286,10 @@ export const simulateMint = async (
         functionName: 'mint',
         args: mintArgs,
         account: from,
-        // TODO: There's a bug in Viem preventing this behavior; log issue with them
-        // stateOverride: [{
```
jonathandiep, line 286: Not totally sure if there was a bug (at least in current version -- 2.9.23) since I was able to get this running fine

Quazia (approved):

Quazia:
@jonathandiep We safe to get this in main? Still looks good to me

jonathandiep:
> @jonathandiep We safe to get this in main? Still looks good to me

I'll get it in today!

PR #406, opened 2024-05-22
chore(neynar): add validateRecast to plugin

sammccord:


Quazia (approved):
Looks good - I checked the interface and the registry too to double check everything else looks connected to me

PR #408, opened 2024-05-22
feat(utils): adds the complete action type

jonathandiep:
Adds the `Complete` Action Type


packages/utils/src/types/actions.ts
```diff
@@ -292,6 +294,26 @@ export const CreateActionDetailSchema = z.object({
 export type CreateActionForm = z.infer<typeof CreateActionFormSchema>
 export type CreateActionDetail = z.infer<typeof CreateActionDetailSchema>
 
+/*
+COMPLETE
+*/
+export type CompleteActionParams = {
+  boostId?: string
```
mmackz, line 301: I think we will also want to take in the chainId as an actionParam.

mmackz: Im not sure if we can make it work for all networks at the same time?

packages/utils/src/types/actions.ts
```diff
@@ -292,6 +294,26 @@ export const CreateActionDetailSchema = z.object({
...
+  actionType?: string
+}
+
+export const CompleteActionFormSchema = z.object({})
```
mmackz, line 305: The form schema is for the inputs we want to take in on manager, (at least this is how I understand it)

So we are going to want to have optional fields here for boostId and actionType

packages/utils/src/types/actions.ts
```diff
@@ -292,6 +294,26 @@ export const CreateActionDetailSchema = z.object({
...
+export const CompleteActionDetailSchema = z
+  .object({
+    boostId: z.string().optional(),
+    actionType: z.string().optional(),
+  })
+  .refine((data) => data.boostId != null || data.actionType != null, {
```
mmackz, line 311: We want the `chainId` in here also (assuming we want to add it to actionParams)

packages/utils/src/types/actions.ts
jonathandiep: I added `chainId` as optional. It should allow us to enable for all networks. Will update if we need to enforce a `chainId`

mmackz (approved):
Looks good 🚢

PR #421, opened 2024-05-29
chore: mark third party sdks as external in build

sammccord:
up until now, we've been compiling third party sdk's as a part of our builds AND specifying them as production dependencies, which means downstream consumers are still downloading the package even though they aren't imported in the published bundle.

this change preserves imports/requires for those sdk's in our build artifacts, thereby deferring to downstream bundlers as far as how to best resolve those dependencies

the end result is faster build times, less memory overhead in ci, and smaller bundles


Quazia (approved):
helllll yeah 🥇

PR #423, opened 2024-05-30
fix: boost plugin

jonathandiep:
Include the tx hash and chain id to be a part of the completion payload


Quazia (approved):
LGTM - were the claim signatures failing because the tx hash wasn't saved in DB?

PR #436, opened 2024-06-04
feat(sound): use `creatorAddress` as referral 

mmackz:
- Use boost `creatorAddress` as dynamic referral address for embedded mints.

- Use ZORA deployer address as a fallback if no creatorAddress is found

### DIFF
--- a/packages/soundxyz/src/constants.ts
+++ b/packages/soundxyz/src/constants.ts
@@ -316,3 +316,7 @@ export const MINT_INFO_LIST_ABI = [
     type: 'function',
   },
 ]
+
+// for referrals
+export const ZORA_DEPLOYER_ADDRESS =
+  '0xe3bBA2A4F8E0F5C32EF5097F988a4d88075C8B48'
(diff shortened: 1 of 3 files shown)


packages/soundxyz/src/constants.ts
```diff
@@ -316,3 +316,7 @@ export const MINT_INFO_LIST_ABI = [
     type: 'function',
   },
 ]
+
+// for referrals
+export const ZORA_DEPLOYER_ADDRESS =
+  '0xe3bBA2A4F8E0F5C32EF5097F988a4d88075C8B48'
```
jonathandiep, line 322: What is the `ZORA_DEPLOYER_ADDRESS`?

jonathandiep (approved):

packages/soundxyz/src/constants.ts
mmackz: Ive been using it as the default referral address for Boost. It is used as a fallback here if no creatorAddress is provided.


It specifically an EOA that we currently use to collect referral income from Zora, Pods and Foundation.

mmackz: Will likely come back and update this once we get some clarity on where we want the referral income to go.

PR #438, opened 2024-06-06
feat(lens): add support for Lens Plugin

mmackz:
## Integration of `lens` collect action plugin for QuestDK

This PR introduces the integration of the `lens` collect plugin for QuestDK.

- [x] [Merge Utils/SDK types for collect action](https://github.com/rabbitholegg/questdk-plugins/pull/439)

### About The Project

Lens Protocol is a decentralized social graph protocol designed to empower developers to build and innovate on top of a user-owned social network, providing seamless integration and interoperability across various applications.

### Implementation Details

Lens Collect action will use the Indexer V4 along with the Lens GraphQL endpoint to determine if a certain address has collected a post.

There is no currently way to query the time that a post was "collected" in Lens using their api, so if someone has collected a post before a boost has started, they will be eligible for rewards. The only way around this would be to capture the data in on-chain events.

The endpoint to query for collected posts is `whoActedOnPublication` which returns a paginated response of addresses who have collected a specific postId.

https://lens-protocol.github.io/lens-sdk/classes/_lens_protocol_client.Core.Profile.html#whoActedOnPublication

There is no way to filter this response, so we check each page to see if the address has collected the post.


packages/lens/vite.config.js
```diff
@@ -0,0 +1,19 @@
+/** @type {import('vite').UserConfig} */
+export default {
+  build: {
+    rollupOptions: {
+      external: [/@rabbitholegg/],
```
mmackz, line 5: do I need to add anything here for external dependancies (ie: apollo)?

packages/lens/package.json
```diff
@@ -0,0 +1,36 @@
...
+    "test": "vitest dev",
+    "test:cov": "vitest dev --coverage",
+    "test:ci": "CI=true vitest --coverage",
+    "test:ui": "vitest dev --ui"
+  },
+  "keywords": [],
+  "author": "",
+  "license": "ISC",
+  "dependencies": {
+    "@apollo/client": "^3.10.4",
```
mmackz, line 31: should I pin?

packages/lens/vite.config.js
jonathandiep: I'm curious about this too @sammccord

packages/lens/src/graphql.ts
```diff
@@ -0,0 +1,71 @@
+import { ActionVariables, PaginatedResponse, Profile } from './types'
+import { ApolloClient, DocumentNode, InMemoryCache, gql } from '@apollo/client'
+
+const APIURL = 'https://api-v2.lens.dev/'
```
jonathandiep, line 4: [nit] add underscore (e.g. `API_URL`)

jonathandiep (approved):

PR #444, opened 2024-06-10
feat(thirdweb): add support for ThirdWeb `mint` plugin

mmackz:
## Integration of `thirdweb` mint action plugin for QuestDK

This PR introduces the integration of the `thirdweb` mint plugin for QuestDK.

### About The Project

Thirdweb is a robust platform that streamlines the process of minting and managing NFTs. By providing easy-to-use tools and infrastructure, it empowers creators to mint, deploy, and manage their NFT collections efficiently. 

### Implementation Details

Thirdweb is available on most major L2 networks such as base and optimism

We have support for two types of contracts, `DropERC1155` and `OpenEditionERC721`, both are quite similar, with the difference being the ERC1155 can utilize a tokenId parameter. Both use the `claim` functions on their respective contracts.

### Sample Mints

  - ERC721: https://wallet.coinbase.com/nft/mint/MDP
  - ERC1155: https://basescan.org/token/0x5625e0ae98c035407258d6752703fed917417add

### Example Transactions

  - ERC721: https://basescan.org/tx/0x16ed42066507832a5452dd1c012b1c6f694284987abcd2d8b3d2e0cdc7eb9678
  - ERC1155: https://basescan.org/tx/0x2f0bba62facc408f510a1ca52e9e41f09e067646854ceda1c4e379caca6c946d


packages/thirdweb/src/ThirdWeb.test.ts
```diff
@@ -0,0 +1,267 @@
+import { mint } from './ThirdWeb'
+// import { getFees, mint, simulateMint } from './ThirdWeb'
```
Quazia, line 2: cleanup

packages/thirdweb/src/utils.ts
```diff
@@ -0,0 +1,102 @@
...
+
+      if (supportsInterface) {
+        return type as '1155' | '721'
+      }
+    } catch {}
+  }
+  throw new Error('Invalid contract type')
+}
+
+export function formatAmount(amount: FilterOperator | undefined) {
```
Quazia, line 52: This and the getMintAmount could both probably live in the utils package.

Quazia (approved):
LGTM
A few helper functions that could be moved into the utils and the `types.ts` file is a little confusing - it looks like mocked returns and you could cast them to the Viem type for that simulation but it doesn't really look like they're types.

packages/thirdweb/src/ThirdWeb.test.ts
mmackz: Ive left the commented code so that I can easily run the non-mocked tests when needed.  (the mocked tests are kind of useless 😅)

Wondering if maybe there is a better way to have non-mocked tests to be run locally.

packages/thirdweb/src/utils.ts
mmackz: Going to make a separate PR to add these functions to utils.

PR #445, opened 2024-06-11
feat(foundation): add support for ERC1155 mints

mmackz:
Refactors the foundation plugin to add support for ERC-1155 mints. 

#### Notes
For 1155 tokens, this implementation requires a tokenId or it will throw an error. Normally we target the initial tokenId in the collection, but this does not work here because the tokenIds are not sequential and can be any random number. 

Here are the relative contracts:

- **Implementation Contract:** https://basescan.org/address/0x6d3e140383606af80b64ec1ee91ed4014ba86d34#code
- **Minter:** https://basescan.org/address/0x132363a3bbf47e06cf642dd18e9173e364546c99 (all base mints happen here)
- **Collection:** https://basescan.org/token/0xddcbe62d10dbee8492610fd1b4964403a0e087aa


packages/foundation/src/utils.ts
```diff
@@ -46,3 +59,101 @@ export function calculateFees(
...
+    return undefined
+  }
+  if (amount && ['string', 'number', 'bigint'].includes(typeof amount)) {
+    return { $gte: amount }
+  }
+
+  return amount
+}
+
+export function getMintAmount(amount: FilterOperator | undefined) {
```
Quazia, line 140: This might actually be worth putting in utils - seems useful for most simulations

Quazia (approved):
This looks good - but as the code exists it's going to cause issues with the way the backend implements fees.

[Here's where it gets the fees](https://github.com/rabbitholegg/rh/blob/31d8c900b2001dd924866194e51839bba8bc1011/src/services/internal/action-params.ts#L35C11-L35C15)

[That value then gets saved into the quest object in the DB](https://github.com/rabbitholegg/rh/blob/31d8c900b2001dd924866194e51839bba8bc1011/src/services/internal/action-params.ts#L79-L85)

Reading through this code the Dutch Auction fees are going to change, so the front-end embedded mint is going to need to look up the fees pre-flight, not rely on the cached values. Potentially requires backend and front-end changes to support the dynamic fees.

mmackz:
> This looks good - but as the code exists it's going to cause issues with the way the backend implements fees.
> 
> [Here's where it gets the fees](https://github.com/rabbitholegg/rh/blob/31d8c900b2001dd924866194e51839bba8bc1011/src/services/internal/action-params.ts#L35C11-L35C15)
> 
> [That value then gets saved into the quest object in the DB](https://github.com/rabbitholegg/rh/blob/31d8c900b2001dd924866194e51839bba8bc1011/src/services/internal/action-params.ts#L79-L85)
> 
> Reading through this code the Dutch Auction fees are going to change, so the front-end embedded mint is going to need to look up the fees pre-flight, not rely on the cached values. Potentially requires backend and front-end changes to support the dynamic fees.

Thankfully the dutch auction fee type is very rarely used. Good callout, the fees will definitely increase after a few mints are made.

What is the best way to deal with this in the short term? Should we just throw an error and prevent people from launching boosts with dutch auction?

Quazia:
I'd say prevent it for now until we can modify it in the front-end and back-end if it's not used high frequency.

PR #448, opened 2024-06-11
fix(pods): fix simulation errors when no tokenId is provided [BOOST-4097]

mmackz:
Previously this would receive an error when no tokenId was provided due to their not being a certain function `nextTokenId` not being available on the ABI.

There can still be issues if the latest tokenId is not in a mintable state.

fixes BOOST-4097

- change tests to use mocks


Quazia (approved):
LGTM - it _is_ getting the most recent token ID now. As a follow-on if we see more fails I still think it would make sense to iterate through the tokenIds when the simulate fails and tokenId hasn't been set but we can pull that lever based on subsequent issues SG?

packages/pods/src/Pods.test.ts
```diff
@@ -1,11 +1,14 @@
...
   type MintIntentParams,
 } from '@rabbitholegg/questdk-plugin-utils'
 import { apply } from '@rabbitholegg/questdk/filter'
+import { mintResponseTokenId, mintResponseUndefined } from './mockedValues'
 import { type Address, parseEther } from 'viem'
-import { describe, expect, test } from 'vitest'
-import { getFees, getMintIntent, mint, simulateMint } from './Pods'
+import { describe, expect, test, vi } from 'vitest'
+import { getMintIntent, mint } from './Pods'
+// import { getFees, simulateMint } from './Pods'
```
Quazia, line 11: Cleanup

packages/pods/src/Pods.test.ts
```diff
@@ -94,15 +97,28 @@ describe('Given the getFee function', () => {
...
+        actionFee: parseEther('0.0007'),
+      }),
+    }
+    const getFeesSpy = vi.spyOn(mockFns, 'getFees')
+    const fee = await mockFns.getFees(mintParams)
+    expect(getFeesSpy).toHaveBeenCalledWith(mintParams)
+    expect(fee.projectFee).toEqual(parseEther('0'))
+    expect(fee.actionFee).toEqual(parseEther('0.0007'))
+
+    // const { actionFee, projectFee } = await getFees(mintParams)
```
Quazia, line 119: Cleanup

packages/pods/src/Pods.test.ts
```diff
@@ -116,11 +132,70 @@ describe('simulateMint function', () => {
...
+      value,
+      address,
+    )
+    expect(simulateMintSpy).toHaveBeenCalledWith(
+      mint as MintIntentParams,
+      value,
+      address,
+    )
+
+    // const result = await simulateMint(mint, value, address)
```
Quazia, line 157: cleanup

packages/pods/src/mockedValues.ts
```diff
@@ -0,0 +1,37 @@
+export const mintResponseUndefined = {
```
Quazia, line 1: I think viem exports the type that that simulation returns

Quazia: `SimulateContractReturnType`

PR #463, opened 2024-06-25
feat(lens): update validation method

mmackz:
- updates validation method to use NFT apis

~After checking through several api providers, the best suited one is blockspan, as it has the needed lookup for our usecase, and it has generous rate-limits available. It does require an api key, so we will need to set up an environment variable.~ 

Choose to use Alchemy SDK instead of blockspan due to high latency for the api to index new mints transactions.

 Reservoir api will be used as a backup option in case ~blockspan~ alchemy goes down or gets rate-limited. Reservoir does not require an API key, but adding one would increase the limits.
 
There is minimal latency using this method, with most requests taking between 600-1000ms, which is around the same as the previous validation methods.
 
 
 ~Will need to add the `BLOCKSPAN_API_KEY` to heroku.~
 
 We already have an `ALCHEMY_API_KEY` available, so no new env variables are required.


packages/lens/src/validate.ts
```diff
@@ -49,24 +50,63 @@ export async function getCollectAddress(postId: string) {
...
 }
 
-export async function checkAddressOwnsCollect(
+export async function checkAddressMintedCollect(
+  actor: Address,
+  collectAddress: Address,
+) {
+  try {
+    return await checkMintedUsingBlockspan(actor, collectAddress)
+  } catch {
```
mmackz, line 59: how to deal with error handling here?

Quazia: If the second check fails we should throw `ValidationNotValid`
https://github.com/rabbitholegg/rh/blob/main/src/indexer/v4/consumers/sqs-plugin-action-validator.ts#L154C28-L154C46

packages/lens/src/validate.ts
```diff
@@ -1,13 +1,15 @@
...
-  production,
-  type SimpleCollectOpenActionSettingsFragment,
   type MultirecipientFeeCollectOpenActionSettingsFragment,
+  type SimpleCollectOpenActionSettingsFragment,
+  production,
 } from '@lens-protocol/client'
-import { getClient } from './client'
-import { Chains } from '@rabbitholegg/questdk-plugin-utils'
+import axios from 'axios'
+import dotenv from 'dotenv'
```
Quazia, line 8: Just use `process.env` directly [as done here](https://github.com/rabbitholegg/questdk-plugins/blob/8e57c6e6844943058a072faf5a633f0f777339a5/packages/neynar/src/Neynar.ts#L30) and see if that works.

mmackz: Tried using process.env directly and it did not work. It logs as undefined.

But obviously it has worked for neynar package, so Im a bit confused 🫤

packages/lens/src/validate.ts
mmackz: It should be caught [here](https://github.com/rabbitholegg/questdk-plugins/blob/060ee11f280706163b2925cee6bc90bd31ae608f/packages/lens/src/Lens.ts#L47-L48) and a `ValidationNotValid`. Is that okay or should I move this logic into the `validateCollect` function.

packages/lens/src/validate.ts
jonathandiep: I think it worked in Neynar because there was a check for `process.env`
https://github.com/rabbitholegg/questdk-plugins/blob/8e57c6e6844943058a072faf5a633f0f777339a5/packages/neynar/src/Neynar.ts#L43

mmackz:
@Quazia I got the ENV situation sorted out. It should just work as long as we have the ENV variable in the backend repo. I removed the dotenv as a dependancy in the lens package.

I also added the `ValidationNotValid` error to the utils package and imported it form there. I feel it will be useful to have it in the utils for any other plugins that will use this validation type. wdyt?

mmackz:
Blockscan not reliable. It does not update fast enough. Took roughly 5 minutes for it to pick up the collect

mmackz:
Switched to use Alchemy SDK. It is appears to be extremely fast to pick and index a fresh mint transaction and is quite responsive. Another benefit of using alchemy is that we already have an api key environment variable available.

Pretty happy with the results.

packages/lens/src/alchemy.ts
```diff
@@ -0,0 +1,9 @@
+import { Alchemy, Network } from 'alchemy-sdk'
+
+const settings = {
+  apiKey: process.env.ALCHEMY_API_KEY,
+  network: Network.MATIC_MAINNET,
+  requestTimeout: 2000,
```
mmackz, line 6: I just put an arbitrary value as the requestTimeout. Thought was to let it timeout soonish and hit the reservoir backup.

Not sure if this is too much or too little.

packages/lens/src/validate.ts
```diff
@@ -20,53 +21,96 @@ export async function hasAddressCollectedPost(
...
+  }
+
+  const transfers = await alchemy.nft.getTransfersForOwner(
+    actor,
+    GetTransfersForOwnerTransferType.TO,
+    options,
+  )
+
+  // only count nfts that originate from zeroAddress
+  const mintedNfts = transfers.nfts.filter((nft) => nft.from === zeroAddress)
```
Quazia, line 89: Clever - you know we could actually handle this just with the contract `Transfer` events. Just ship this implementation though that's hilarious it's full circle though.

Quazia (approved):

PR #467, opened 2024-07-10
fix(zora): amount zero fix

mmackz:
- removes legacy writing from abi (it has not quantity)
- fixes issue where zero amount mints can potentially pass the filter
- uses mint amount utilities
- adds additional test for zero amount mint


packages/zora/src/Zora.ts
```diff
@@ -105,8 +107,14 @@ export const mint = async (
...
 
-  const andArray721 = []
-  const andArray1155 = []
+  const initialArray = [
+    {
+      quantity: formatAmount(amount),
+    },
+  ]
+  const andArray721: AndArrayItem[] = [...initialArray]
+  const andArray1155: AndArrayItem[] = [...initialArray]
```
jonathandiep, line 116: [nit] Can convert the `initialArray` into the object so you don't need to spread the array

```
const quantityCheck = {
  quantity: formatAmount(amount),
}
const andArray721: AndArrayItem[] = [quantityCheck]
const andArray1155: AndArrayItem[] = [quantityCheck]

jonathandiep (approved):

packages/zora/src/Zora.ts
mmackz: updated. good catch.

PR #473, opened 2024-07-10
fix(fabric): fix amount zero issue

mmackz:
- add check for num tokens in validation
- fix amount zero issue
- update types
- adjust tests to catch error when no contract data is found


Quazia (approved):
So they have a variable minimum order required?

mmackz:
> So they have a variable minimum order required?

Depends on the minimum time required for a subscription (ie: 1 day, 1 week, etc)

The amount is static, meaning it stays the same for all mints.

PR #479, opened 2024-07-15
feat(pods): use referral in mint validation

mmackz:
- uses `referral` param in mint validation
- uses `referral` from action params in simulate int and mint intents functions
- add tests to verify referral validation is working correctly

Fixes BOOST-4258

https://linear.app/rh-app/issue/BOOST-4258/update-pods-to-use-referral-param-in-mint-validation-function


packages/pods/src/Pods.ts
```diff
@@ -26,20 +26,35 @@ import {
...
     },
   ]
+  if (referral) {
+    const checksumAddress = getAddress(referral)
+    andArray1155.push({
+      $or: [
+        { mintReferral: checksumAddress },
+        {
+          rewardsRecipients: {
+            $and: [{ $first: checksumAddress }, { $last: checksumAddress }],
```
mmackz, line 52: @Quazia We want to make sure that the referral is the only item in the array. Would using an array literal work better here? ie: `[checksumAddress]`

jonathandiep (approved):

packages/pods/src/Pods.ts
mmackz: Used the array literal instead of $and.

PR #481, opened 2024-07-15
feat(zora): use `referral` param in validation and simulations

mmackz:
- use referral param for simulations and validation
- add additional tests

Fixes BOOST-4257

https://linear.app/rh-app/issue/BOOST-4257/update-zora-to-use-referral-param-in-mint-validation-function


packages/zora/src/Zora.ts
```diff
@@ -126,6 +127,17 @@ export const mint = async (
       tokenId,
     })
   }
+  if (referral) {
+    const referralAddress = getAddress(referral)
```
Quazia, line 131: This will always return check summed - I'd add a callout since elsewhere we consistently use lower case

Quazia (approved):
LGTM but same question around wondering if we use define default referral address in utils. Also might be worthwhile to just alias the deployed address so they aren't bound to the same variable but that may be an unnecessary optimization.

mmackz:
> LGTM but same question around wondering if we use define default referral address in utils. Also might be worthwhile to just alias the deployed address so they aren't bound to the same variable but that may be an unnecessary optimization.

I have a ZORA_DEPLOYER_ADDRESS used in several plugins now, so it would make sense to save it as `REFERRAL_ADDRESS` and export it from utils. Ill submit a ticket and make the changes at a later date.

packages/zora/src/Zora.ts
mmackz: Removed this line since using a checksummed address is not required anymore.

I was using an array operator previously for the `rewardsRecipients`, which required a checksummed address.

PR #484, opened 2024-07-15
feat(foundation): add referral param to mint validation

mmackz:
- add referral param to validation, simulation and mint intents functions
- adds additional tests to verify validation works properly with referral param
- check tests using live data (marked ones that are failing w/ comments)

Fixes BOOST-4259

https://linear.app/rh-app/issue/BOOST-4259/update-foundation-to-use-referral-param-in-mint-validation-function


Quazia (approved):
LGTM - do we want a referral address defined at the util level or is it too different per protocol where we want to default referral too? It's okay it's the same for all networks too, right?

Also good callout for the live data on the tests. The auction is disabled right or how does this behave if someone _does_ target an auction?

mmackz:
> LGTM - do we want a referral address defined at the util level or is it too different per protocol where we want to default referral too? It's okay it's the same for all networks too, right?
> 
> Also good callout for the live data on the tests. The auction is disabled right or how does this behave if someone _does_ target an auction?

It is disabled in the code, which causes the live test to fail (as expected)

mmackz:
I talked to foundation, and they told me they have cancelled their referral program for foundation.app. Ive reached back to them for more info on if they will honour any referrals that are embedded in the transactions.

mmackz:
> I talked to foundation, and they told me they have cancelled their referral program for foundation.app. Not sure exactly what this means, since you can still embed a referral directly in the transaction.
> 
> Ive reached back to them for more info on if they will honour any referrals that are embedded in the transactions.

They support referrals at the transaction level, but they have deprecated referrals using url params.

PR #485, opened 2024-07-16
feat(foundation): implement `getExternalUrl` function

mmackz:
- implement `getExternalUrl` function
- add tests

- [x] https://github.com/rabbitholegg/questdk-plugins/pull/478

BOOST-4268

https://linear.app/rh-app/issue/BOOST-4268/implement-getexternalurl-in-plugins-foundation

### DIFF
--- a/packages/foundation/src/Foundation.test.ts
+++ b/packages/foundation/src/Foundation.test.ts
@@ -411,7 +411,9 @@ describe('Given the foundation plugin', () => {
         contractAddress: '0xead6dca70b0465725a57eb81f7d3ab8b5e0b81b4',
       }
       const result = await getExternalUrl(mint)
-      expect(result).toBe('https://foundation.app/mint/base/0xead6dca70b0465725a57eb81f7d3ab8b5e0b81b4')
+      expect(result).toBe(
+        'https://foundation.app/mint/base/0xead6dca70b0465725a57eb81f7d3ab8b5e0b81b4',
+      )
     })
 
     test('should return the base url for 1155 mint', async () => {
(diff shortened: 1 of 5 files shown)


packages/foundation/src/Foundation.test.ts
```diff
@@ -403,4 +403,37 @@ describe('Given the foundation plugin', () => {
...
+    })
+
+    test('should return the base url for 1155 mint', async () => {
+      const mint: MintActionParams = {
+        chainId: Chains.BASE,
+        contractAddress: '0x1d2550d198197df1a10af515cf2ea0d790889b93',
+        tokenId: 213,
+      }
+      const result = await getExternalUrl(mint)
+      expect(result).toBe('https://foundation.app/')
```
jonathandiep, line 426: This test result doesn't look correct? Shouldn't it also include the action params in here?

jonathandiep: Nvm I see that this is intentional. Is it possible to extract the url from Foundation? Or will it be a super deep dive into their API or something?

jonathandiep (approved):
lgtm, but would like to try to get the 1155 external url (if it's low lift!)

packages/foundation/src/Foundation.test.ts
mmackz: There erc-1155 implementation is custom and has no available frontend. The last one was for their babes collection.

https://babes.foundation.app/

mmackz: It might be available in the contract or token uri

mmackz: No luck.

mmackz:
> lgtm, but would like to try to get the 1155 external url (if it's low lift!)

Was not able to find any way to derive an external url for 1155 unfortunately. Will update if something comes up.

PR #496, opened 2024-07-23
feat(titles): add support for Titles `mint` plugin

mmackz:
## Integration of `titles` mint action plugin for QuestDK

This PR introduces the integration of the `titles` mint plugin for QuestDK.

### About The Project

TITLES builds creative tools powered by artist-owned AI models. 

### Implementation Details

The mint plugin for TitlesV2 is pretty straight-forward. All mints will use the `mint` function of the titles V2 Collection Contract.

- **tokenId is required** to generate projectFees and for simulations. If one is not provided it will throw an error.
- There is no referral query params available (afaik), so referral are only available through embedded mints.
- The plugin is only setup for BASE. They have contracts on OP and Zora, but there is no examples to go off. Will add at a later date.

#### Sample Mints
- https://titles.xyz/collect/base/0x432f4ccc39ab8dd8015f590a56244becb8d16933/1

#### Sample Transactions

- https://basescan.org/tx/0xe12969d93f9a8241e031246f295ac69c304e1968b2ea568680cbaf93683030dd


Quazia (approved):
LGTM - we want to get this live ASAP so we can maybe just let them test it out and see if it's working as expected.

It looks like they'll always need to simulate so they won't be able to create a boost without a tokenId - that makes sense right?

PR #500, opened 2024-07-24
feat: add actionType to getExternalUrl

mmackz:
- adds actionType argument to getExternalUrl function

Fixes BOOST-4313


mmackz:
Fixes BOOST-4313

PR #502, opened 2024-08-08
feat(zora): update plugin to be compatible with zora protocol upgrade [BOOST-4411]

mmackz:
- Updates plugin to use the new sales strategy config (0.000111 base fees)

- This will only work for free mints at the moment. Will follow-up with update to get the fee calculations working correctly. Fee calculations requires us to update the Zora-sdk which introduces a whole bunch of breaking changes. 

Everything is working in local environment except validation. I dont think it is a plugin issue as the transaction shows as OK when using the transaction checker tool.

CleanShot 2024-08-08 at 09 29 01@2x


packages/zora/src/Zora.ts
```diff
@@ -363,6 +348,66 @@ export const simulateMint = async (
...
+    client ??
+    (createPublicClient({
+      chain: chainIdToViemChain(mint.chainId),
+      transport: http(),
+    }) as PublicClient)
+
+  try {
+    return await simulateV2Mint(mint, value, _client, account)
+  } catch {
+    return await simulateV1Mint(mint, value, _client, account)
```
jimobrien, line 407: this failover will obscure `simulateV2Mint` errors that aren't caused by the mint being v1 which could make debugging simulation errors more difficult. 

if there's a specific error we can check for to know if the mint is v1, that would be ideal. otherwise logging the error would give visibility into what's going on: 

```
  try {
    return await simulateV2Mint(mint, value, _client, account)
  } catch (error) {
    console.error("simulateV2Mint failed:", error)
    return await simulateV1Mint(mint, value, _client, account)
  }
  ```

File: packages/zora/src/Zora.ts [hunk 2]
```diff
@@ -406,7 +451,8 @@ export const getFees = async (
     return { actionFee: fee.tokenPurchaseCost, projectFee: fee.mintFee }
   } catch (err) {
     console.error(err)
-    return { actionFee: parseEther('0'), projectFee: parseEther('0.000777') } // https://github.com/ourzora/zora-protocol/blob/e9fb5072112b4434cc649c95729f4bd8c6d5e0d0/packages/protocol-sdk/src/apis/chain-constants.ts#L27
+    // ! temp change to base fee of 0.000111 until we update for fee calulation
+    return { actionFee: parseEther('0'), projectFee: parseEther('0.000111') } // https://github.com/ourzora/zora-protocol/blob/e9fb5072112b4434cc649c95729f4bd8c6d5e0d0/packages/protocol-sdk/src/apis/chain-constants.ts#L27
```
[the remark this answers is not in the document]
mmackz: Havent fully looked into it yet, but I assume the updated zora sdk should have a way to deal with this.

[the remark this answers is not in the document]
Quazia: The SDK should provide a solve here - we also do cache this value still I think so the existing boosts shouldn't break and new boosts will fail on the creation simulation.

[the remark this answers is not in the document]
Quazia: Does increase the need for updating the SDK though IMO

packages/zora/src/Zora.ts
```diff
@@ -363,6 +348,67 @@ export const simulateMint = async (
...
+  client: PublicClient,
+  account?: Address,
+): Promise<SimulateContractReturnType> => {
+  const { contractAddress, tokenId, amount, recipient, referral } = mint
+
+  const from = account ?? DEFAULT_ACCOUNT
+
+  const _tokenId = await getNextTokenId(client, contractAddress, tokenId)
+
+  await checkBytecode(client, contractAddress)
```
Quazia, line 363: No need to move this now but I could see this being generally useful - nice little portable helper.

packages/zora/src/contract-addresses.ts
```diff
@@ -12,3 +12,5 @@ export const FIXED_PRICE_SALE_STRATS: { [chainId: number]: Address } = {
 }
 
 export const ZORA_1155_FACTORY = '0x777777c338d93e2c7adf08d102d45ca7cc4ed021'
+export const ZORA_TIMED_SALE_STRATEGY =
+  '0x777777722d078c97c6ad07d9f36801e653e356ae'
```
Quazia, line 16: It's nice they're actually deterministically deploying this now

packages/zora/src/Zora.ts
```diff
@@ -179,10 +174,39 @@ export const mint = async (
...
+  }
+}
+
+const v2MintFilter = (mint: MintActionParams) => {
+  const { chainId, contractAddress, tokenId, amount, recipient, referral } =
+    mint
+
+  return {
+    chainId,
+    to: getExitAddresses(chainId, ZORA_TIMED_SALE_STRATEGY),
```
Quazia, line 190: It does look like it's still possible to go through `0x777777722D078c97c6ad07d9f36801e653E356Ae` in addition to `0x777777722D078c97c6ad07d9f36801e653E356Ae` and the actual collection.

Quazia (changes requested):
I'm seeing mints get routed through the new strategy and the [Bulk Mint Tx](https://zorascan.xyz/tx/0x32f5f85ee662625f8d3f22b1f662f010a4525163cfaccf221c504f6e0e5c9883) - ZoraMintsManager

I poked around and didnt' see any instances of mints for this strat getting fired against the collection contract directly but I reckon it's safer to add that to the exist addresses short term.

Outside of that some minor comments and nits but the biggest thing to re-investigate is the exist addresses.

packages/zora/src/Zora.ts
Quazia: Hmmmmmmmm - it looks like this address is `ZoraTimedSaleStrategy` on OP and `ZoraMintsManager` on Zora?

Quazia (approved):

PR #507, opened 2024-08-09
feat(zora): update zora SDK and implement getFees

mmackz:
- Updates zora SDK to latest version
- Updates method for getting `salesConfig` using updated SDK.
- Implements `getFees` function for V2 mints
- Removes `getMintIntent` tests. They are failing when not mocked due to the mins no longer being active.
- Removes Blast as a supported chain (not compatible with updated SDK)

Fixes BOOST-4415


mmackz:
Fixes BOOST-4415

mmackz:
https://linear.app/rh-app/issue/BOOST-4415/update-zora-sdk-and-fix-getfees-calculations

Quazia (approved):

PR #511, opened 2024-08-15
fix(zora): add tokenId to zora SDK getToken call for 1155s

mmackz:
- Updated Zora SDK requires a tokenId on the `getToken` call, otherwise it will fail and return the base fee.

Fixes BOOST-4477

https://linear.app/rh-app/issue/BOOST-4477/zora-getfees-fails-for-1155-wo-tokenid-plugin

### DIFF
--- a/packages/zora/src/Zora.ts
+++ b/packages/zora/src/Zora.ts
@@ -456,7 +456,8 @@ export const getFees = async (
   mint: MintActionParams,
 ): Promise<{ actionFee: bigint; projectFee: bigint }> => {
   try {
-    const { chainId, contractAddress, tokenId, amount } = mint
+    const { chainId, contractAddress, amount } = mint
+    let tokenId = mint.tokenId
 
     const client = getPublicClient(chainId)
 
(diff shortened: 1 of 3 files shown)


packages/zora/src/Zora.ts
```diff
@@ -471,6 +472,12 @@ export const getFees = async (
       publicClient: client,
     })
 
+    // if contract type is 1155, we need to get the latest tokenId
+    if (contractType === '1155' && tokenId == null) {
```
Quazia, line 476: if we pass tokenId into `getNextTokenId` it should account for the existing `tokenId` right? Neutral on if we pass it or implementation, important thing is that the simulation and fees are both using the same logic to determine the token to use (both use next if tokenId is undefined)

Quazia (approved):
LGTM

packages/zora/src/Zora.ts
mmackz: getNextTokenId passes back the existing tokenId if it is defined. 

Felt this was more semantic.

SimulateMint and getFees will both use `getNextTokenId` when no token is available.

PR #512, opened 2024-08-15
fix(titles): fix failing simulate mint test

mmackz:
Test is failing due to the target NFT not being mintable.

Switched the test to one that targets an NFT that is open indefinitely.

https://linear.app/rh-app/issue/BOOST-4478/fixtitles-fix-failing-simulate-mint-test

Fixes BOOST-4478


Quazia:
Fixes [Boost-4478]

Quazia (approved):

PR #514, opened 2024-08-16
Implement Zod schemas for Filters in utils [BOOST-4470]

sammccord:
- Implement Zod schemas for Filters in utils
- Fix several explicit any's to resolve ts issues
- Update Mint params schema to match number or FilterOperator
- Update zod to support readonly usage in abitype

### DIFF
--- a/packages/utils/src/types/filters.ts
+++ b/packages/utils/src/types/filters.ts
@@ -1,26 +1,9 @@
+// eslint-disable-next-line import/no-internal-modules
+import { Abi as AbiSchema } from 'abitype/zod'
 import type { Abi, Transaction } from 'viem'
+import { z } from 'zod'
 
-export type ArrayOperator =
-  | {
-      $some?: FilterOperator[]
-    }
-  | {
-      $first?: FilterOperator
-    }
-  | {
-      $last?: FilterOperator
-    }
-  | {
-      $nth?: NthFilter
-    }
-
-export type LogicalOperator =
-  | {
-      $and?: FilterOperator[]
-    }
-  | {
-      $or?: FilterOperator[]
-    }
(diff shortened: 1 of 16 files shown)


tsconfig.json
```diff
@@ -12,6 +12,7 @@
     "noImplicitOverride": true,
     "noUnusedLocals": true,
     "noUnusedParameters": true,
+    "noImplicitAny": false,
```
sammccord, line 15: this was giving me guff trying to build, figured that at this point any any's we have right now are fine and we're unlikely to introduce a great deal more core functionality

packages/utils/src/types/filters.ts
```diff
@@ -39,41 +22,175 @@ export type NumericOperator =
...
+  .object({
+    $abiParams: AbiSchema,
+  })
+  .catchall(z.lazy(() => FilterSchema))
+
+export type Filter =
+  | Primitive
+  | FilterObject
+  | FilterArray
+  | FilterOperator
```
sammccord, line 176: @Quazia is this correct? I added the `FilterOperator` to `Filter` union type, and schema. Unsure if it's the move.

Quazia: I think that is correct - some of the issues before may have been from _not_ having it in this.

Quazia (approved):
Looks good - this must have been a tough one 😅 

Just for my edification why'd you end up casting the context as any when passing it through in the handlers?

sammccord:
> Looks good - this must have been a tough one 😅
> 
> Just for my edification why'd you end up casting the context as any when passing it through in the handlers?

The context type unions don't always overlap with the operator type unions, especially when it comes to `Record<string, unknown>` vs transaction interfaces.

PR #517, opened 2024-08-21
[BOOST-4470] Further expand filter interfaces, schemas

sammccord:
- use `Filter` as base type for mint action params, supports any combination of:
  - `FilterOperator`
  - `FilterObject`
  - `FilterArray`
  - `Primitive`
  - `PrimitiveArray`
- values of `FilterObject` and `FilterOperator` are now `Filter`, which is not a breaking change because `Filter` includes `FilterOperator`
- support wider range of filter combinations like
```{
  // Logical filter operator  
  "$or": [  
    {
      // FilterObject  
      "to": {
        // Logical filter operator  
        "$or": [
          // Primitive Array  
          "0xfoo",  
          "0xbar"  
        ]  
      },
```

### DIFF
--- a/.changeset/dry-ducks-sleep.md
+++ b/.changeset/dry-ducks-sleep.md
@@ -1,0 +1,5 @@
+---
+"@rabbitholegg/questdk-plugin-utils": minor
+---
+
+allow any combination of filters in mint params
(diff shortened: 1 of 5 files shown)


jonathandiep (approved):
lgtm

